PR #1, opened 2015-10-09
Add support for client-side batching to generate fewer <style> tags.

jlfwong:
This also adds some examples to the repo, and paves the path forward for serverside rendering.

Test Plan:

```
cd examples
npm install
npm run examples
```

Open http://localhost:4114/, see 2 style tags generated: one for initial render, one for the dynamic blue added by the timer.

@xymostech

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,11 +65,55 @@ Support for colocating your styles with your React component.
         }
     });
 
+# Buffering
+
+To avoid making a new style tag for each individual call to `css`, you can use 
+buffering. A similar technique will enable server rendering.
+
+On the client, instead of just this:
+
+    ReactDOM.render(<App/>, document.getElementById('root'));
+
+You can do this:
+
+    StyleSheet.startBuffering();
+    ReactDOM.render(<App/>,
+                    document.getElementById('root'),
+                    StyleSheet.flush);
+
+Note that this is an optimization, and the first option will work just fine.
+
+Once implemented, server rendering will look roughly like this:
+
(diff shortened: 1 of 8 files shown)


src/index.js
```diff
@@ -29,6 +33,32 @@ const StyleSheet = {
...
+            return;
+        }
+        if (injectionBuffer.length > 0) {
+            injectStyles(injectionBuffer);
+        }
+        injectionMode = 'IMMEDIATE';
+        injectionBuffer = "";
+    },
+
+    css(styleDefinitions) {
```
xymostech: This can be removed, correct? (since it's now merged into `css` below?)

README.md
```diff
@@ -80,6 +124,7 @@ Support for colocating your styles with your React component.
   the unit is. See
   [CSSProperty.js](https://github.com/facebook/react/blob/master/src/renderers/dom/shared/CSSProperty.js)
   in React.
+- Consider removing !important from everything.
```
xymostech: :+1: having everything be `!important` is pretty annoying with dynamic styles, because if you want a plain `style={...}` style to overwrite an aphrodite style then you need to append `" !important"` to it.

xymostech:
This all looks fine to me (except for the function to be removed)!

src/index.js
jlfwong: Whoops, yep -- bad merge

README.md
zgotsch: I too have felt this pain.

PR #5, opened 2015-10-18
💖Emoji Separators 💖

jlfwong:
It turns out I was wrong! CSS selectors are totally allowed to contain unicode characters!

http://www.w3.org/TR/CSS21/syndata.html#tokenization

image

If we wanted to be more byte-frugal, we could also use `\:` or `\+`, since escape sequences are also allowed.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -44,7 +44,7 @@ const css = (function() {
             return "";
         }
 
-        const className = validDefinitions.map(s => s._name).join("-o_O-");
+        const className = validDefinitions.map(s => s._name).join("\u{1F496}");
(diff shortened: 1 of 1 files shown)


rileyjshaw:
💖

zgotsch:
This surprisingly even works in ie9

xymostech:
I'd vote for something non-emoji (despite them being great). Mostly because emoji don't show up in my browser. What does `\:` or `\+` look like when you inspect it?

jlfwong:
Will re-investigate later using `\:` or `\+`, and see if that has reasonable browser support.

PR #7, opened 2015-10-21
Add browserify settings to package.json -_-

zgotsch:
This is gross and I don't like it, but Browserify doesn't transform node_modules by default. We use Browserify at Spring and this lib needs to run through Babel since it's ES6. I think it's ok to add this to `package.json`, since it won't affect non-Browserify users and it will make the lib work with browserify, so long as the user has the Babelify transform.

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,5 +27,8 @@
     "jsdom": "^6.5.1",
     "mocha": "^2.3.3",
     "webpack": "^1.12.2"
+  },
+  "browserify": {
+    "transform": ["babelify"]
   }
 }


jaredly:
I'd much rather setup a `prepublish` hook `babel -d lib/ src` and have the `index.js` that `package.json` points to use the `lib` transpiled stuff. That makes this lib available to people regardless of transpilation. Best practice w/ webpack is also to exclude `node_modules` from transpilation.

jlfwong:
Agree with @jaredly

zgotsch:
That would also be awesome.

jlfwong:
Fixed in https://github.com/Khan/aphrodite/commit/6719bb11bc5c92c4db5c3ff16cb3c14c5eae5292, just need to wait on @xymostech to publish to npm (we camped the npm package while this was still a private repo, but now it's public, so we should publish it for real).

zgotsch:
You da best

PR #11, opened 2015-11-04
Add a propType for aphrodite styles.

xymostech:
Summary: Since we recommend people pass around aphrodite styles instead of
calling `css` and then passing that around, we should provide a good proptype
to validate that. This adds that proptype, and some tests to make sure it
works.

Test plan:
 - `npm run test`

Reviewers: @jlfwong


jlfwong:
This needs rebasing, but we should otherwise land this (sorry for the huge delay)

xymostech:
This isn't up to date any more, closing for now.

kentcdodds:
Also, on this note, even though the original use case is for use with React, that's not a requirement and aphrodite works great with other things as well. So I'd recommend this be implemented in a separate npm module and a reference added to the README. That way people who aren't using React don't have to pay for the bytes.

jlfwong:
The other benefit of it being a separate npm package is that it can depend on react directly, and make use of other `React.PropTypes`.

kentcdodds:
> it can depend on react directly

Yup! Though, to be clear, it should list React as a (very semver-loose) `peerDependency` so it doesn't cause people to have multiple versions of React in their app in some cases (when using npm@2 for example). :+1:

PR #12, opened 2015-11-04
Add a `combine` function for combining multiple styles into one.

xymostech:
Summary: This allows people to pass around a single style instead of having to
deal with arrays of styles and `...rest`ing them into calls to `css`. Having
the `css` function handle this seems a bit funky because then we'll end up
generating styles for things that aren't used, but a bunch of the code is
duplicated between these two functions already.

Do you think it would make sense for `css` to do this combining, or should I
maybe make `css` call out to the `combine` function so there isn't duplicated
code?

Test plan: `npm run test`

Reviewers: @jlfwong

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -1,5 +1,5 @@
 import {generateCSS} from './generate';
-import {mapObj, nextID} from './util';
+import {mapObj, nextID, flatten, recursiveMerge} from './util';
 
 const injectStyles = (cssContents) => {
     // Taken from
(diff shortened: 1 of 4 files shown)


jaredly:
Does this change the public interface at all? fwiw, in React native, you can do `<View style={[arbitrarily, [nested, arrays], of, styles]}>`
It also ignores falsy values.
Might be useful to mirror functionality as much as possible

jlfwong:
I like the idea of having `css` flatten, then this shouldn't be an issue

PR #13, opened 2016-01-05
Make it so that !important can be disabled for certain style rules.

jeresig:
Specifically disable it for all @font-face rules as it breaks them. cc: @cjfuller

### DIFF
--- a/src/generate.js
+++ b/src/generate.js
@@ -2,7 +2,8 @@ import {
     objectToPairs, kebabifyStyleName, recursiveMerge, stringifyValue
 } from './util';
 
-export const generateCSS = (selector, styleTypes, stringHandlers) => {
+export const generateCSS = (selector, styleTypes, stringHandlers,
+        useImportant) => {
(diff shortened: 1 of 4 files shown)


jeresig:
cc'ing @jlfwong + @xymostech

src/generate.js
```diff
@@ -20,24 +21,27 @@ export const generateCSS = (selector, styleTypes, stringHandlers) => {
...
     );
 };
 
-export const generateCSSRuleset = (selector, declarations, stringHandlers) => {
+export const generateCSSRuleset = (selector, declarations, stringHandlers,
+        useImportant) => {
     const rules = objectToPairs(declarations).map(([key, value]) => {
         const stringValue = stringifyValue(key, value, stringHandlers);
-        return `${kebabifyStyleName(key)}:${stringValue} !important;`;
+        const important = (useImportant === false ? "" : " !important");
```
xymostech: nit: Maybe move this outside of the `map`? Premature optimization or something...

xymostech:
This looks great to me!

PR #19, opened 2016-02-12
An example of better messaging with the README

### LINKED ISSUE #18: This works without React.
While I was reading the README, I was thinking that nothing about this library has any dependence/lock-in with React. Although it pairs wonderfully, the beginning of the README says:

> Support for colocating your styles with your React component.

Which implies it only works with React. I found this isn't the case, with my demo using Custom Elements and template strings here:

https://gist.github.com/tbranyen/4336b59e11812fb679d3

Would it make sense to follow the Redux approach ("Redux is a predictable state container for JavaScript apps."), and not focus on React?

tbranyen:
I complained about the README wording in a GitHub issue and this is my
attempt of rewording the prominent React messaging.

Fixes #18

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,8 +2,9 @@
 
 [![npm version](https://badge.fury.io/js/aphrodite.svg)](https://badge.fury.io/js/aphrodite)
 
-Support for colocating your styles with your React component.
+Support for colocating your styles with your JavaScript component.
 
+- Works great with and without React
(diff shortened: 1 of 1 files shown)


jlfwong:
Thanks for your contribution!

tbranyen:
Thanks for helping me through it @jlfwong. Love this project can't wait to see it evolve.

PR #20, opened 2016-02-13
Updated webpack.config to generate umd build instead of commonjs

furqanZafar:
Updated webpack.config to generate a umd build & added a makefile to generate both dist/aphrodite.js & dist/aphrodite.min.js. 

once published to npm the umd build can be accessed via npmcdn  (https://npmcdn.com/aphrodite@0.1.3/dist/aphrodite.js & https://npmcdn.com/aphrodite@0.1.3/dist/aphrodite.min.js)

Example (a prefix free pen): http://codepen.io/anon/pen/wMNywy


jlfwong:
Thanks for the PR!

I think I'd prefer this build rule to be part of the `"script"` section of `package.json`. The dev version is already created use `npm run build`. I can imagine the other one could be made via `npm run build:min`.

At KA, we're using the commonjs version as distributed right now, and would like to continue doing so without the unnecessary bytes created by the UMD version. That said, I'm happy to *also* have a UMD version checked into the dist folder for use via the CDN, perhaps as `dist/aphrodite.umd.js` and `dist/aphrodite.umd.min.js`.

If you do wish to pursue this route, I'd like for `npm run build` to build all three files (`dist/aphrodite.js`, `dist/aphrodite.umd.js` and `dist/aphrodite.umd.min.js`) to avoid getting into a situation where we accidentally update one but not the others.

Would you be willing to handle making those updates?

furqanZafar:
Sure no problem. :+1:

xymostech:
Hey @furqanZafar, any updates on progress for this?

xymostech:
I'm gonna close this in favor of #55

PR #21, opened 2016-02-17
improved server-side rendering example code

montemishkin:
Just some minor fixes to the documentation for server-side rendering.

- fixed a typo (`css.contents` -> `css.content`)
- added `$`s for ES6 string template syntax
- added necessary call to `JSON.stringify`

Thanks for the great package!


jlfwong:
Thanks for the all the fixes!

PR #22, opened 2016-02-18
remove need for `StyleSheet.create`

montemishkin:
It seems to me that the `StyleSheet.create` api is unnecessary.  

Removing it would (among other things) simplify the api, and minimize the hassle for users considering moving over from libraries like [radium](https://github.com/FormidableLabs/radium) for example, where there are no calls to `StyleSheet.create` as far as the library user is concerned.

So far, this PR only contains the rather small changes to `src/index.js` needed to make things work without the call to `StyleSheet.create`.  

I'm happy to fix up the (currently failing) test suite etc., but figured I would run the idea by you all before getting too far ahead.  After all, perhaps I'm missing some bigger picture, and `StyleSheet.create` is actually a good thing.

But it seems like it's totally not needed.

One thing to note is that these changes make it more difficult to track down the generated CSS class from your JS styles object (since the generated class name is no longer prefixed by the JS styles object key).

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -6,17 +6,6 @@ import {
 } from './inject';
 
 const StyleSheet = {
-    create(sheetDefinition) {
-        return mapObj(sheetDefinition, ([key, val]) => {
-            return [key, {
-                // TODO(emily): Make a 'production' mode which doesn't prepend
-                // the class name here, to make the generated CSS smaller.
-                _name: `${key}_${hashObject(val)}`,
-                _definition: val
-            }];
-        });
-    },
-
     rehydrate(renderedClassNames=[]) {
         addRenderedClassNames(renderedClassNames);
     },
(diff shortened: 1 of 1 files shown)


jlfwong:
Thanks for the thoughtful PR description!

We would like to maintain the API of `StyleSheet.create` and not have people pass arbitrary object properties into the `css` function. Doing so allows us to maintain API stability while changing the API considerably.

Aside from that higher level design goal, this PR would result in us hashing the object in every call to `css` instead of once per `StyleSheet.create` call, which would likely contribute significantly to runtime (I'd have to profile to be sure). This could be solved by memoizing hashes of objects, but I'd rather keep the `StyleSheet.create`.

We have no specific goal of making it easy to migrate from similar libraries such as radium. The `StyleSheet.create` API is meant to closely mimic the API for CSS in React Native: https://facebook.github.io/react-native/docs/stylesheet.html

montemishkin:
Three very good points.  Thanks for taking a look :)

rxb:
Sorry for the noise, but just wanted to say thanks for choosing this approach. It made Aphrodite exactly the choice I needed. 🙌

PR #27, opened 2016-03-15
Upgrade to Babel 6 and remove default from export

adrianleb:
Since the recent code changes when adding prefix styles, when running `npm prepublish` Babel was crashing on this line https://github.com/Khan/aphrodite/blob/master/src/util.js#L29 since it expects a plugin to be able to transpile said code.

As a consequence, the build published to NPM failed to contain the latest `/lib` files, if you right now install version `0.1.2` of aphrodite via NPM it will not be the latest according to what's on Github although the package.json version number indicates so (one good hint is the outdated README.md and complete lack of auto-prefixing code which is what made me investigate this problem)

With this PR I've added the necessary Babel plugin and a .babelrc file, so running `npm run prepublish` works as expected. the last step is to publish a new version to NPM so everyone has, for sure, the latest version of aphrodite including prefixed styles magic!

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,8 +9,8 @@
   ],
   "main": "lib/index.js",
   "scripts": {
-    "test": "mocha --compilers js:babel/register tests",
-    "test:watch": "mocha --watch --compilers js:babel/register tests",
+    "test": "mocha --compilers js:babel-core/register tests",
+    "test:watch": "mocha --watch --compilers js:babel-core/register tests",
     "build": "webpack --config webpack.config.js",
     "build:watch": "webpack --watch --config webpack.config.js",
     "prepublish": "babel -d lib/ src"
@@ -26,9 +26,12 @@
   },
   "homepage": "https://github.com/Khan/aphrodite",
   "devDependencies": {
-    "babel": "^5.8.23",
-    "babel-core": "^5.8.25",
-    "babel-loader": "^5.3.2",
+    "babel": "^6.5.2",
+    "babel-cli": "^6.6.5",
+    "babel-core": "^6.7.2",
+    "babel-loader": "^6.2.4",
+    "babel-plugin-transform-object-rest-spread": "^6.6.5",
+    "babel-preset-es2015": "^6.6.0",
     "chai": "^3.3.0",
     "jsdom": "^6.5.1",
     "mocha": "^2.3.3",
--- a/.babelrc
+++ b/.babelrc
@@ -1,0 +1,6 @@
+{
+    "presets": ["es2015"],
+    "plugins": [
+        "transform-object-rest-spread",
+    ]
+}
--- a/src/index.js
+++ b/src/index.js
@@ -56,7 +56,7 @@ const css = (...styleDefinitions) => {
     return className;
 };
 
-export default {
+export {
     StyleSheet,
     StyleSheetServer,
     css,


kevinbarabash:
The travis-ci build is failing probably because it's trying to use a babel 6 plugin with babel 5.  We should probably upgrade to babel 6.

adrianleb:
Right, I can upgrade to Babel 6 part of this PR so that travis-ci is happy.

Any reason to keep using Babel 5? I don't see any non-trivial change required.

kevinbarabash:
If you'd be willing to upgrade to Babel 6 that would be awesome.  If you haven't done so already, please sign our Contributor License Agreement: http://www.khanacademy.org/r/cla.

adrianleb:
CLA Signed ✅
Babel 6 upgrade ✅
Travis now passing ✅
One small change to be noted, on https://github.com/Khan/aphrodite/pull/27/files#diff-1fdf421c05c1140f6d71444ea2b27638R59 I've removed the `default` since the compiled result was not what was expected (and reflecting the docs) This article covers why http://www.2ality.com/2015/12/babel-commonjs.html#why_isnt_the_default_export_done_like_a_commonjs_single_export%3F

kevinbarabash:
Looks good.  I'm going let @xymostech or @jlfwong hit the merge button because they are the main contributors for this repo.

.gitignore
```diff
@@ -1,2 +1,3 @@
 lib
 node_modules
+.DS_Store
```
xymostech: This should be in a global `.gitignore` and not be added here.

xymostech:
This all looks good to me @adrianleb! Good call on the `default` usage, I'm guessing this works for us because we just use the ES5 compiled version.

Could you squash your commits?

jlfwong:
Hi @adrianleb, thanks for the PR!

I'm a little confused about the scenario you originally ran into, however. The `npm prepublish` script should've used the version in `node_modules`, which should've been Babel 5 per the existing `package.json`. That script doesn't crash with Babel 5, right? So how was it using Babel 6 for you?

adrianleb:
Hey @jlfwong, no problem, my pleasure!

So the scenario is that when using `0.1.2` from NPM the code !== to what's on Github as 0.1.2, you can check the difference here:
https://raw.githubusercontent.com/Khan/aphrodite/master/dist/aphrodite.js
vs
https://npmcdn.com/aphrodite@0.1.2/dist/aphrodite.js

Notably, this means everything related to inline-prefix-styles.

After checking out master and trying to prepublish locally babel was getting stuck on the line I mentioned above (https://github.com/Khan/aphrodite/blob/master/src/util.js#L29)

I do have babel installed globally at version `6.5.1`

If prepublish works with the current version of babel and `{...restSpread}` out of the box on a machine different than mine then the fix to this issue is just another call to `npm publish` with the right code for `0.1.2`

jlfwong:
@adrianleb Thanks for the explanation -- the difference between what's on GitHub and what's on npm isn't surprising. It sounds like we should version bump and publish to put the style prefixing up on npm.

I'm still a little confused about the global babel installation -- I would've thought that npm scripts (e.g. if you had run `npm prepublish`) would set a PATH that forces the `node_modules/.bin` to take precedence over a global installation. I guess that might not be the case if you hadn't first run `npm install`, because the `node_modules/.bin` would be empty then?

If I understand correctly, this PR shouldn't be fixing any bugs, but just upgrading to Babel 6 (which is then, IMO, what this PR should be titled with). Please let me know if I'm still misundertanding

adrianleb:
@jlfwong You're absolutely correct!

I've tested again on a different computer and the version on master runs fine, I guess I had a bad conf on my machine.

In short, this PR becomes about upgrading to latest Babel & removing the `default` from the main export line which is not needed/not the best pattern.

Most important, a polite request to have the NPM version of aphrodite updated since that's what triggered me to do this. :)

I've updated the PR title and squashed the commits so If still fitting the code is ready to be merged.

jlfwong:
@adrianleb Thanks for investigating on another computer :)

I've published 0.2.0 now, which should contain the style prefixing changes. Thanks for letting us know it was out of date.

As for this diff, I'm not super inclined to jump on upgrading without cause at the moment -- other portions of Khan Academy's internal code uses babel 5 at the moment, so I'd rather leave this alone unless there are pressing needs.

When we do get around to upgrading though, this will serve as a great reference for other repositories, so I'm going to leave this open for now if that's alright with you.

adrianleb:
Thanks @jlfwong!

Agree updating Babel has way less urgency if everything works.

The one thing which this PR also cover which can be worth looking into is the `default` export.

Right now to use Aphrodite I'm having to do `import Aphrodite from 'aphrodite';` since according to recent es6 changes `import {Stylesheet, css} from 'aphrodite'` doesn't work if there's only a `default` export.

It's not exactly a bug but documentation can be a bit misleading about it.

jlfwong:
@adrianleb Oh, interesting! If Babel 5 will take the version without the `default`, then I'm happy to take that as a PR by itself.

package.json
```diff
@@ -26,9 +26,12 @@
   },
   "homepage": "https://github.com/Khan/aphrodite",
   "devDependencies": {
-    "babel": "^5.8.23",
-    "babel-core": "^5.8.25",
-    "babel-loader": "^5.3.2",
+    "babel": "^6.5.2",
```
lencioni: FWIW, I ended up doing a similar update in https://github.com/Khan/aphrodite/pull/281

kevinbarabash:
We've update to using babel 6 internally quite some time ago.

kevinbarabash:
It looks like Aphrodite was upgraded to use babel 6 as well some time ago so I'm going to close this.

PR #34, opened 2016-03-27
Add keyframes support

xymostech:
Summary: This adds support for keyframes using the same mechanism that
we support `font-family` with. By passing in an object of keyframes as
the value of `animationName`, this will generate an `@keyframes` block
and return the name.

Test Plan:
 - `npm run test`
 - Visit the examples page at http://localhost:4114/, see that the "This
   should animate" text animates.
 - Visit the ssr examples page at http://localhost:4114/ssr,
   see the text still animates.

@jeresig @jlfwong

### DIFF
--- a/src/inject.js
+++ b/src/inject.js
@@ -1,6 +1,7 @@
 import asap from 'asap';
 
 import {generateCSS} from './generate';
+import {hashObject} from './util';
 
 // The current <style> tag we are inserting into, or null if we haven't
 // inserted anything yet. We could find this each time using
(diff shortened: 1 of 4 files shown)


src/inject.js
```diff
@@ -69,11 +113,8 @@ let injectionBuffer = "";
 // already be flushed, or because we are statically buffering on the server.
 let isBuffering = false;
 
-export const injectStyleOnce = (key, selector, definitions, useImportant) => {
+const injectGeneratedCSSOnce = (key, generatedCSS) => {
     if (!alreadyInjected[key]) {
```
jeresig: Is this still needed if it's also in `injectStyleOnce`?

src/inject.js
```diff
@@ -55,6 +56,49 @@ const stringHandlers = {
...
+    //      top: 5,
+    //    },
+    //    to: {
+    //      left: 20,
+    //      top: 20,
+    //    }
+    //  }
+    // TODO(emily): figure out a way to call this `animation` instead of
+    // `animationName`.
+    animationName: (val) => {
```
jeresig: Nice! Glad you were able to make use of the `stringHandlers` functionality.

src/inject.js
```diff
@@ -55,6 +56,49 @@ const stringHandlers = {
...
+        if (typeof val === "object") {
```
jeresig: To tighten up the code a bit you could do:

```
if (typeof val !== "object") {
    return val;
}
```

and then just have the rest of the code (helps to save on an extra indentation).

src/inject.js
jeresig: Instead of `animationName` could you have it be `keyframes` maybe?

src/inject.js
```diff
@@ -55,6 +56,49 @@ const stringHandlers = {
...
+            // Generate a unique name based on the hash of the object. We can't
+            // just use the hash because the name can't start with a number.
+            // TODO(emily): this probably makes debugging hard, allow a custom
+            // name?
+            const name = `keyframe_${hashObject(val)}`;
+
+            // Since keyframes need 3 lays of nesting, we use `generateCSS` to
```
jeresig: *layers

src/inject.js
```diff
@@ -55,6 +56,49 @@ const stringHandlers = {
...
+            // build the inner layers and wrap it in `@keyframes` ourselves.
```
jeresig: I'm kind of confused as to why the existing CSS generation logic doesn't work here. Shouldn't the inner layers just be generated like any other CSS sub-class?

jeresig:
@xymostech Great work! I've added some comments inline.

src/inject.js
xymostech: It doesn't work, because aphrodite stringifies the values of the second layer, instead of generating more css. So, `a: { '0%': { left: 0 } }` generates `a { 0%: [object Object]; }` instead of `a { 0% { left: 0px; } }` like we want.

src/inject.js
xymostech: Well, the CSS key name needs to be `animation-name` in order for it to actually use the animation, and stringHandlers doesn't let us rename what the key is, only change the value. I guess my TODO should be more clear about what doesn't work.

jeresig:
Great, this all looks good to me! Thank you for the follow-ups :)

jlfwong:
Add to README, please!

PR #39, opened 2016-04-06
	Export StyleSheet.create as StyleSheet function

ajoslin:
Less typing! :clap: 

ES5: 

```js
var sheet = require('aphrodite').StyleSheet

module.exports = sheet({
  foo: {background: 'red'}
})
```

ES6:

```js
import {StyleSheet} from 'aphrodite'

export default StyleSheet({
  foo: {background: 'red'}
})
```

This is a non-breaking change, all of the other exports still exist.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -5,22 +5,24 @@ import {
     addRenderedClassNames, getRenderedClassNames
 } from './inject';
 
-const StyleSheet = {
-    create(sheetDefinition) {
-        return mapObj(sheetDefinition, ([key, val]) => {
-            return [key, {
-                // TODO(emily): Make a 'production' mode which doesn't prepend
-                // the class name here, to make the generated CSS smaller.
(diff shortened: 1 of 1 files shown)


xymostech:
We'd like to keep the creation function as `StyleSheet.create` to keep compatibility with other libraries like [React Native](https://facebook.github.io/react-native/docs/style.html), [React look](https://github.com/rofrischmann/react-look), etc.

ajoslin:
This *does* keep compatibility, it only adds a shortcut. It's a non-breaking change, all other exports still exist.

jlfwong:
IMO, it's better to have exactly one way of doing things. Syntactic sugar is nice and all, but it creates more confusing documentation.

If you'd like this functionality for your own projects, it should be easy enough to make a wrapper around Aphrodite that provides this sugar.

ajoslin:
Understood :+1:

PR #40, opened 2016-04-07
Updated syntax by removing unnecessary comma from API styles object

brukh:
+ Updated syntax by removing unnecessary comma from API styles object


jlfwong:
Ah, this trailing comma was intentional. Part of Khan Academy's style guide.

PR #44, opened 2016-04-09
use triple-equal consistently

sarbbottam:
I guess `!styleTag` should be fine.


spicyj:
`x == null` is intentional shorthand for `x === null || x === undefined`.

sarbbottam:
Thanks! Shouldn't `!styleTag` serve the purpose? Anyhow, could there be a comment for stating the intention to avoid PRs?

PR #46, opened 2016-04-10
using object shorthand

sarbbottam:
Similar to [aphrodite/src/index.js#L59-L63](https://github.com/Khan/aphrodite/blob/master/src/index.js#L59-L63)
```js
export default {
    StyleSheet,
    StyleSheetServer,
    css,
};
```

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -27,14 +27,14 @@ const StyleSheetServer = {
         reset();
         startBuffering();
         const html = renderFunc();
-        const cssContent = flushToString();
+        const css = {
+          content: flushToString(),
+          renderedClassNames: getRenderedClassNames(),
+        }
 
         return {
-            html: html,
-            css: {
-                content: cssContent,
-                renderedClassNames: getRenderedClassNames(),
-            },
+            html,
+            css,
         };
     },
 };


xymostech:
I think this obscures what the structure of the return value of this function is.

sarbbottam:
Does this sound better?

```diff
-const html = renderFunc();
-const cssContent = flushToString();

return {
-    html: html,
+    html: renderFunc(),
    css: {
-        content: cssContent,
+        content: flushToString(),
        renderedClassNames: getRenderedClassNames(),
    },
};
```

May be, I don't follow the rational of caching the content of `html` and `css.content` in a local variable, and not for `renderedClassNames`. Any specific reason?

xymostech:
We need to ensure that `renderFunc()` is called before `flushToString()` or `getRenderedClassNames()`, which is why it's separate (I don't know if it's specified which order things will execute). I can't remember if it's necessary that `flushToString()` comes before `getRenderedClassNames()`, but it might be?

sarbbottam:
Ahh, I see, thanks for the clarification, I guess it would be better, to add these comments in-line, to avoid unnecessary PRs and ... 

> I can't remember if it's necessary ...

PR #49, opened 2016-04-10
removed unnecessary `map` call

sarbbottam:
I guess the map call is `unnecessary` here, please close the PR if I have misinterpreted.

### DIFF
--- a/src/inject.js
+++ b/src/inject.js
@@ -48,7 +48,7 @@ const stringHandlers = {
     // an array of objects and strings.
     fontFamily: function fontFamily(val) {
         if (Array.isArray(val)) {
-            return val.map(fontFamily).join(",");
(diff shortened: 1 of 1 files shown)


xymostech:
This map is necessary, it recursively calls `fontFamily` on each element of the `fontFamily` array.

sarbbottam:
Thanks @xymostech ! I completely misinterpreted it. Did not encounter a test failure, thus was in doubt.

xymostech:
Yeah, I don't think we have any tests for the `fontFamily` string handler (there's a good place for it here: https://github.com/Khan/aphrodite/blob/master/tests/inject_test.js#L206)

sarbbottam:
I can take it up.

PR #50, opened 2016-04-10
use consistent regex naming convention

sarbbottam:
it would be better to use a consistent naming convention


xymostech:
Can you move the `IMPORTANT_RE` declaration back to above `importantify` since that's the only place it's used?

sarbbottam:
updated, thanks!

PR #53, opened 2016-04-12
enable non-commonjs consumers

kentcdodds:
By choosing `umd` as the `libraryTarget`, people can use `aphrodite` without a bundler which is great for adoption. Note: people can continue using `CommonJS`. This is not a breaking change :-)

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -7,7 +7,8 @@ module.exports = {
   output: {
     path: path.join(__dirname, 'dist'),
     filename: 'aphrodite.js',
-    libraryTarget: "commonjs2"
+    libraryTarget: 'umd',
+    library: 'aphrodite'
   },
   module: {
     loaders: [{


xymostech:
There's already a pull request to do this, #20, but they haven't updated their pull request with the feedback. I'll go ping them. If they don't respond, would you mind making the same changes @jlfwong suggested?

> At KA, we're using the commonjs version as distributed right now, and would like to continue doing so without the unnecessary bytes created by the UMD version. That said, I'm happy to also have a UMD version checked into the dist folder for use via the CDN, perhaps as dist/aphrodite.umd.js and dist/aphrodite.umd.min.js.

> If you do wish to pursue this route, I'd like for npm run build to build all three files (dist/aphrodite.js, dist/aphrodite.umd.js and dist/aphrodite.umd.min.js) to avoid getting into a situation where we accidentally update one but not the others.

kentcdodds:
Definitely willing to do that if #20 doesn't update :+1:

kentcdodds:
It [looks like](https://github.com/Khan/aphrodite/blob/fe27e13d026f867110faeea5ece85ed0532824bc/package.json#L16) you're using babel to transpile the code to CommonJS. Is there a reason that you need webpack at all to support KA's use case of using commonjs? In fact [the `main`](https://github.com/Khan/aphrodite/blob/fe27e13d026f867110faeea5ece85ed0532824bc/package.json#L10) is actually pointing to the babel version anyway, so you're not even using the webpack version at all. I'm going to submit another PR here in a second to address some other things with scripts, but I don't think that we need to do anything special with webpack to support anything but the UMD build and we'll just use babel for the CommonJS.

spicyj:
KA's build system doesn't support many parts of commonjs so tends to use browser builds of things instead of installing directly from npm.

kentcdodds:
Ah, now that's interesting... Ok, I'll make sure that I support both. Thanks for that!

kentcdodds:
Created a whole new PR because it's a bit different: https://github.com/Khan/aphrodite/pull/55

PR #48, opened 2016-04-10
added missing unitless numbers

sarbbottam:
in sync with [React's CSSProperty.js](https://github.com/facebook/react/blob/master/src/renderers/dom/shared/CSSProperty.js#L17-L54)

PR #54, opened 2016-04-12
move to inline-style-prefix-all

natew:
This breaks two tests, which I've opened a related ticket here:

https://github.com/rofrischmann/inline-style-prefix-all/issues/6

### DIFF
--- a/dist/aphrodite.js
+++ b/dist/aphrodite.js
@@ -218,6 +218,9 @@ module.exports =
 	 */
 	var isUnitlessNumber = {
 	    animationIterationCount: true,
+	    borderImageOutset: true,
+	    borderImageSlice: true,
+	    borderImageWidth: true,
 	    boxFlex: true,
 	    boxFlexGroup: true,
 	    boxOrdinalGroup: true,
@@ -243,8 +246,11 @@ module.exports =
 
 	    // SVG-related properties
 	    fillOpacity: true,
+	    floodOpacity: true,
 	    stopOpacity: true,
+	    strokeDasharray: true,
 	    strokeDashoffset: true,
+	    strokeMiterlimit: true,
 	    strokeOpacity: true,
 	    strokeWidth: true
 	};
@@ -350,13 +356,13 @@ module.exports =
 	};
 
 	exports.hashObject = hashObject;
-	var importantRegexp = /^([^:]+:.*?)( !important)?$/;
+	var IMPORTANT_RE = /^([^:]+:.*?)( !important)?$/;
 
 	// Given a style string like "a: b; c: d;", adds !important to each of the
 	// properties to generate "a: b !important; c: d !important;".
 	var importantify = function importantify(string) {
 	    return string.split(";").map(function (str) {
-	        return str.replace(importantRegexp, function (_, base, important) {
+	        return str.replace(IMPORTANT_RE, function (_, base, important) {
 	            return base + " !important";
 	        });
 	    }).join(";");
--- a/package.json
+++ b/package.json
@@ -36,6 +36,6 @@
   },
   "dependencies": {
     "asap": "^2.0.3",
-    "inline-style-prefix-all": "^1.0.4"
+    "inline-style-prefix-all": "^1.0.5"
   }
 }
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "motion-aphrodite",
-  "version": "0.2.0",
+  "version": "0.2.1",
   "description": "Inline styles in JS that just work (TM)",
   "keywords": [
     "css",
--- a/package.json
+++ b/package.json
@@ -1,5 +1,5 @@
 {
-  "name": "aphrodite",
+  "name": "motion-aphrodite",
   "version": "0.2.0",
   "description": "Inline styles in JS that just work (TM)",
   "keywords": [
@@ -36,6 +36,6 @@
   },
   "dependencies": {
     "asap": "^2.0.3",
-    "inline-style-prefixer": "Khan/inline-style-prefixer#f41f3040ac27eeec3b7a1fb7450ddce250cac4e4"
+    "inline-style-prefix-all": "^1.0.4"
   }
 }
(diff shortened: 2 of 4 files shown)


natew:
For background: the version of inline-style-prefixer that aphrodite uses actually causes bugs in packagers because it has both `Plugins/index.js` and `plugins.js` and it expects to require the former. I'm not sure on the spec here actually, so perhaps this is fine.

package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "aphrodite",
+  "name": "motion-aphrodite",
```
steelbrain: I think you might want to revert this change

jlfwong:
Thanks for the PR! Wasn't aware that repo was created.

Can't merge this until that issue you mentioned is fixed, but would certainly be nice to not maintain our own fork.

Would you mind posting the gzipped file size before and after this change? I'd just like to make sure this doesn't increase the file size dramatically.

@xymostech Any idea if there are any nice pre-built bots that would report the gzipped file size on the diff? It's going to be a question for pretty much any major PR: how many bytes does it add?

xymostech:
@jlfwong no idea! Looking around it doesn't seem like it. :( I guess theoretically we should ask people to check in the built files, and we can just look at the diff there?

@natew: I think that bug is one of the things I fixed in our branch of inline-style-prefixer that I never got around to pushing back upstream... I fixed it here: https://github.com/Khan/inline-style-prefixer/blob/fa1fc3732a84fa760b03cc6afd09d7da4340bd85/modules/plugins/transition.js

natew:
@jlfwong Not sure the total diff but inline-style-prefixer-all has a pretty prominent disclaimer of it's size (inline-style-prefix-all is a tiny (3.3kb gzipped)) on the repo.

jlfwong:
@natew With the version bump, the tests are failing. I think the decision made in https://github.com/rofrischmann/inline-style-prefix-all/issues/6 is pretty reasonable, so we can adjust our tests to reflect that behavior. That would mean making a breaking change though. Since we're still 0.x.x, I think that's okay? I would be a little surprising if people were relying upon the behavior to expand one vendor prefix into all relevant vendor prefixes.

As for the size change, you should be able to  evaluate it by running something roughly like:

```
git checkout master
npm run build
uglifyjs dist/aphrodite.js | gzip | wc -c
```

then repeating those steps on your branch and comparing the numbers

natew:
master - 7504
this branch - 6658

Using inline-style-prefixer-all 1.0.5

jlfwong:
> master - 7504
> this branch - 6658

Awww yiss. That's the right direction :) Thanks @natew and @rofrischmann!

This should be good to merge after the test fixes.

natew:
@jlfwong changed the two tests to cover the different functionality (just translate prefixes directly don't add others). Let me know if that looks ok!

dist/aphrodite.js
```diff
@@ -218,6 +218,9 @@ module.exports =
 	 */
 	var isUnitlessNumber = {
 	    animationIterationCount: true,
+	    borderImageOutset: true,
```
xymostech: These look like duplicate work of #48 ?

tests/generate_test.js
```diff
@@ -29,24 +29,18 @@ describe('generateCSSRuleset', () => {
         }, '.foo{background-color:red !important;}');
     });
 
-    it('prefixes vendor props with a dash', () => {
+    it('prefixes transition to webkit', () => {
```
xymostech: This test name should be left the same. It was (originally) testing to make sure that our `WebkitTransition` to `-webkit-transition` conversion works.

Why does this only produce `-webkit-transition` prefixes now, and no `-ms-transition` or `-moz-transition`?

rofrischmann: @xymostech `-ms-transition` actually never existed. `-moz-` was left out as it was only used up to Firefox v15 (~2012).

xymostech: Cool! Can you change the name of the test back? Other than that, this all looks good!

dist/aphrodite.js
natew: I'm not sure why this is showing in the diff, I hadn't touched these areas!

tests/generate_test.js
natew: Reverted in latest

dist/aphrodite.js
kentcdodds: It's because you ran the build and committed the dist directory. This is why I recommend not committing generated files: “Why I don’t commit generated files to master” @kentcdodds https://medium.com/@kentcdodds/why-i-don-t-commit-generated-files-to-master-a4d76382564

natew: Oh yea I agree. I typically trust the gitignore to handle that though but I can revert that unless you find time too.

kentcdodds: Want correcting you. Mostly just a suggestion to the maintainers of this project :-)

natew:
Any word on this branch? Once it merges we can delete our fork :smile:

xymostech:
I'm fine with this! Looks like @jlfwong said he's okay with it after the test were fixed, which they are!

Do you need us to publish a new npm version after the merge? We should get in the habit of doing that more regularly...

kentcdodds:
> We should get in the habit of doing that more regularly...

I'm just going to leave this here: npm.im/semantic-release ;-)

xymostech:
@kentcdodds That looks really interesting! It doesn't solve the problem of us failing to `npm publish` very often, though ;) (at KA we use a submodule).

kentcdodds:
Not sure I understand. It effectively allows you to release with every merge to master automatically. (This is configurable, but the automated releases is awesome).

xymostech:
Oh woah. I'm apparently bad at reading and somehow missed that it was automated. That looks really cool!

kentcdodds:
For a quick video tutorial, check out the relevant bit from my free Egghead course: http://kcd.im/write-oss :+1:

0xR:
A quick analysis of the bundle size using source-map-explorer brings inline-style-prefix-all at 9371 bytes minified, not gzipped. Whereas aphrodite is at 7056 bytes minified not gzipped.

PR #55, opened 2016-04-14
improve npm scripts and add more builds

kentcdodds:
Here's how things work now with this PR:

1. There are 4 builds:
  - `main` uses babel and is what the package.json main points to. This
    is pretty much the same as before
  - `umd` uses webpack and specifies that the output should use UMD and
    it outputs a sourcemap file as well.
  - `umdmin` is the same as `umd` except it's run in webpack production
    mode (minified)
  - `commonjs` is exactly as it was before, a bundled module exported as
    a CommonJS module
2. There is a `watch` script associated with each of these
3. There is a `build` script which runs all the builds in parallel
4. There is a `watch:build` script which runs all the builds in watch
   mode in parallel
5. The `prepublish` script was removed because it's not doing what you
   probably think it's doing: (https://github.com/npm/npm/issues/3059)
   and we now have a `release` script which runs the build first and
   then runs `npm publish`. Just run `npm run release` instead of
   `npm publish`.
6. `.gitattributes` now excludes everything in `dist` from `diff`
   because we now have several files exporting there

That's it! Here's a gif showing this in action:

aphrodite

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,9 +11,18 @@
   "scripts": {
     "test": "mocha --compilers js:babel/register tests",
     "test:watch": "mocha --watch --compilers js:babel/register tests",
-    "build": "webpack --config webpack.config.js",
-    "build:watch": "webpack --watch --config webpack.config.js",
-    "prepublish": "babel -d lib/ src"
+    "prebuild": "rimraf dist lib",
+    "build": "npm-run-all --parallel build:*",
+    "watch:build": "npm-run-all --parallel watch:build:*",
+    "build:main": "babel -d lib/ src",
+    "watch:build:main": "npm run build:main -- --watch",
+    "build:umd": "webpack --output-library-target umd --output-library aphrodite --output-filename aphrodite.umd.js --devtool source-map",
+    "watch:build:umd": "npm run build:umd -- --watch",
+    "build:umdmin": "webpack --output-library-target umd --output-library aphrodite --output-filename aphrodite.umd.min.js -p --devtool source-map",
+    "watch:build:umdmin": "npm run build:umdmin -- --watch",
+    "build:commonjs": "webpack --output-library-target commonjs2 --output-filename aphrodite.js",
+    "watch:build:commonjs": "npm run build:commonjs -- --watch",
+    "release": "npm run build && npm publish"
   },
   "repository": {
     "type": "git",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -11,9 +11,18 @@
...
     "test:watch": "mocha --watch --compilers js:babel/register tests",
-    "build": "webpack --config webpack.config.js",
-    "build:watch": "webpack --watch --config webpack.config.js",
-    "prepublish": "babel -d lib/ src"
+    "prebuild": "rimraf dist lib",
+    "build": "npm-run-all --parallel build:*",
+    "watch:build": "npm-run-all --parallel watch:build:*",
+    "build:main": "babel -d lib/ src",
+    "watch:build:main": "npm run build:main -- --watch",
+    "build:umd": "webpack --output-library-target umd --output-library aphrodite --output-filename aphordite.umd.js --devtool source-map",
```
jlfwong: Typo: aphordite

kentcdodds: Nice catch. Updated!

jlfwong:
https://github.com/npm/npm/issues/3059 is pretty sad, yeah. It's a little scary to remove it and provide an avenue to publish a package with out of date ES5 sources (i.e. if I were to run `npm publish` after this diff is landed instead of `npm release`). I guess there isn't a better solution at the moment though?

Pinging @xymostech to ensure you see the change to the prepublish to be release instead.

kentcdodds:
You could always try using [`semantic-release`](http://npm.im/semantic-release). Then you don't need to worry about releases at all. I use it on most of my libraries and it works really well.

kentcdodds:
Note to clarify. I'm sure you know that it's actually `npm run release` not simply `npm release` :-)

xymostech:
Everything looks good to me, the only oddity I see is there isn't a source map for dist/aphrodite.js. Not sure it's necessary, just pointing it out! Any thoughts?

kentcdodds:
Could definitely do that, but I don't think it'd make sense in a normal common js environment...

xymostech:
I guess that's right! Okay, gonna merge this.

jlfwong:
@kentcdodds waaay late question here, but what's the rationale behind the UMD version using a sourcemap? Having sourcemaps checked into the repo doesn't sit super well with me. Going to gitignore the sourcemaps for now, but please let me know if you think we should do otherwise.

jlfwong:
I'm also a little confused by the workflow proposed by `release`.

Here's the intuitive thing to me for publishing:

1. `npm version [major|minor|patch]`
2. npm publish

Where `npm version` does a git commit for us. If `release` does both the build and the publish, it would have to be used after `npm version`, but then the commit with the version bump has already happened!

If I'm understanding right, we could put the build in the `version` script, as described here: https://docs.npmjs.com/cli/version, am I missing something?

PR #56, opened 2016-04-14
Added a caveats section to README.md

conorfennell:


README.md
```diff
@@ -173,6 +173,33 @@ const styles = StyleSheet.create({
 
 Aphrodite will ensure that the global `@font-face` rule for this font is only inserted once, no matter how many times it's referenced.
 
+### Caveats
```
kentcdodds: I think that the Heading that should be used here is probably equal to the other headings, so a single `#` is probably good.

README.md
```diff
@@ -173,6 +173,33 @@ const styles = StyleSheet.create({
...
+
+#### Assigning a string to a content property for a pseudo-element
```
kentcdodds: Likewise, here should probably be `##`

README.md
```diff
@@ -173,6 +173,33 @@ const styles = StyleSheet.create({
...
+
+When assigning a string to the `content` property it requires double quotes in CSS.
```
kentcdodds: This is not necessarily true. Because this would work also:

```javascript
  large: {
      ':after': {
        content: "'Aphrodite'",
      },
    },
});
```

As would: 


```javascript
  large: {
      ':after': {
        content: `'Aphrodite'`,
      },
    },
});
```

What matters is that the value is a string which evaluates to a quoted string to be assigned to the `content` css property. Not sure of the best way to word that though...

kentcdodds:
Other than some nits, I say this LGTM (but my vote doesn't really matter as this isn't really _my_ project) :-)

README.md
conorfennell: True, made the changes to the text to reflect this and added in an extra example.

README.md
```diff
@@ -173,6 +173,42 @@ const styles = StyleSheet.create({
 
 Aphrodite will ensure that the global `@font-face` rule for this font is only inserted once, no matter how many times it's referenced.
 
+# Caveats
+
+## Assigning a string to a content property for a pseudo-element
+
+When assigning a string to the `content` property it requires double or single quotes in CSS.
+Therefore with Aphrodite you also have to provide the quotes within the value string for `content` to match how it will be represented in CSS.
```
kentcdodds: Great way to say this :+1:

README.md
conorfennell: Changed to `##`

README.md
conorfennell: Changed to `#`

xymostech:
Awesome! This looks great!

PR #61, opened 2016-04-20
Add support for descendant selectors.

### LINKED ISSUE #10: How to handle descendant selection of pseudo-states?
If I want to apply styles to a child based on its parent's `:hover` state, I think I have to track hover state in js. Is that accurate? Do you think handling this is in the scope of aphrodite?

Specifically I am thinking of something that emulates the CSS:
```
.parent:hover .child {
    background-color: blue;
}
```

xymostech:
Summary: This adds support for doing descendant style selection; that is,
changing child component styles based on the state/existence of a parent. This
selection in particularly useful when combined with the use of pseudo-classes,
by generating styles that look like `.parent:hover .child`, so the child styles
can depend on the active state of the parent.

Fixes #10

An example of the syntax that is added in this commit (and an example for the
explanation in this commit message):

```js
    parent: {
        '>>child': {
            color: "red"
        },
        ':hover': {
            '>>child': {
                color: "blue"
            },
            '>>otherchild': {
                color: "white"
            }
        },
    },

    altparent: {
        ':hover': {
            '>>child': {
                color: "green"
            }
        }
    }
});
```

It turns out that the API for aphrodite doesn't play well with descendent
styles. In particular, it's difficult to let *descendants have unique
classnames* for themselves, at the same time as allowing *merging between
styles which contain descendant selectors*.

This pull request attempts to do both of these things. The code is a bit messy,
so I'll lay out what's going on in this pull request. Please ask questions, if
something is confusing! The Aphrodite code is nice because it's all fairly
easily understood, and I'm afraid that this adding too much complexity.

The basic flow of this diff is:

1. In `StyleSheet.create`, we recurse through the passed-in styles, and find
each of the descendant selectors (which have keys that look like `>>blah`). In
the example above, it would find `parent['>>child']`,
`parent[':hover']['>>child']`, `parent[':hover']['>>otherchild']`, and
`altparent[':hover']['>>child']`. In each place, we:
  - generate a class name for that descendant selector. This is based on the
    class name of the parent class, as well as the key name.

    For example, if the class name for `styles.parent` was `parent_abcdef`, we
    might generate the class name `parent_abcdef__child` for `parent['>>child']`.

  - tag the style by adding a `_names` object, with the class name as a key of
    the object.

    For example, `parent['>>child']` would end up looking like `{ color: "red",
    _names: { parent_abcdef__child: true } }`.

  - collect a map of each of the keys (without the `>>` bit) to their class
    names.

    For example, for `styles.parent`, we would generate a map that looks like
    `{ child: "parent_abcdef__child", otherchild: "parent_abcdef__otherchild"
    }`.
We merge in the map from key to class name into the generated style, so that
the class names can be accessed using a syntax like `styles.parent.child`.

2. When *parent* styles are passed into `css()`, their styles are merged
together. If one style overrides another's descendant styling, the `_names`
object will be merged together and will contain all of the associated class
names.  
For example, when evaluating `css(styles.parent, styles.altparent)`, we would
end up with merged styles looking like:
    ```
    {
        '>>child': {
            color: "red",
            _names: { parent_abcdef__child: true },
        },
        ':hover': {
            '>>child': {
                color: "green",
                _names: {
                    parent_abcdef__child: true,
                    altparent_123456__child: true,
                },
            },
            '>>otherchild': {
                color: "white",
                _names: { parent_abcdef__otherchild: true },
            }
        }
    }
    ```
We then generate a map from the descendent keys to all of the class names that
could be associated with a given key by recursing and looking at each of the
`_names` objects. For example, the map would look like:
    ```
    {
        '>>child': ["parent_abcdef__child", "altparent_123456__child"],
        '>>otherchild': ["parent_abcdef__otherchild"]
    }
    ```
When generating the styles, we look at this map and then generate styles for
each of the classnames listed. This is so that these styles will match up with
uses of both `css(styles.parent.child)` and `css(styles.altparent.child)`.
For example, when generating the `style[':hover']['>>child']` styles, we
generate:
    ```
    .parent_abcdef-o_O-altparent_123456:hover .parent_abcdef__child { ... }
    .parent_abcdef-o_O-altparent_123456:hover .altparent_123456__child { ... }
    ```

3. When *descendant* styles are passed into `css()`, like
`css(styles.parent.child)`, we simply return the associated class name (in this
case, `"parent_abcdef__child"`) in the output.

Test Plan:
 - `npm run test`
 - `cd examples && npm run examples`, then visit http://localhost:4114/ and see
   that the last line starts green and when "Hover over me" is hovered, the
   other part turns blue.

@zgotsch @jlfwong @kentcdodds @montemishkin

### DIFF
--- a/README.md
+++ b/README.md
@@ -173,6 +173,45 @@ const styles = StyleSheet.create({
 
 Aphrodite will ensure that the global `@font-face` rule for this font is only inserted once, no matter how many times it's referenced.
 
+# Descendant selectors
+
+Descendant rules (i.e. a rule like `.parent:hover .child { ... }` in CSS) are created by adding values to your styles indicating the styles to apply to the child. To distinguish the styles from normal CSS keys, the keys start with a `>>`, like `'>>child'`.
+
+```js
+const styles = StyleSheet.create({
+    someContainer: {
+        ":hover": {
+            ">>myChild": {
+                // Styles to apply to the child when the parent is
+                // hovered over.
+                color: "red",
+                fontWeight: "bold"
+            }
+        }
+    }
+});
+```
+
+Then, apply the `someContainer` style using `css(styles.someContainer)`, and add the child's style to a descendant of the parent using `css(styles.someContainer.myChild)`, where `myChild` is the name of the key you used, without the `>>`.
+
+For example, using React's jsx syntax:
+
+```jsx
+<div className={css(styles.someContainer, ...)}>
+    <div className={css(styles.someContainer.myChild, ...)}>
+        This turns red and bold when the parent is hovered over
+    </div>
+</div>
+```
+
+This will generate CSS that looks like:
+```css
+.someContainer_xxx:hover .myChild_xxx {
+    color: red;
+    font-weight: bold;
+}
+```
+
 # Caveats
 
 ## Assigning a string to a content property for a pseudo-element
--- a/dist/aphrodite.js
+++ b/dist/aphrodite.js
@@ -60,9 +60,47 @@ module.exports =
 
 	var _slicedToArray = (function () { function sliceIterator(arr, i) { var _arr = []; var _n = true; var _d = false; var _e = undefined; try { for (var _i = arr[Symbol.iterator](), _s; !(_n = (_s = _i.next()).done); _n = true) { _arr.push(_s.value); if (i && _arr.length === i) break; } } catch (err) { _d = true; _e = err; } finally { try { if (!_n && _i['return']) _i['return'](); } finally { if (_d) throw _e; } } return _arr; } return function (arr, i) { if (Array.isArray(arr)) { return arr; } else if (Symbol.iterator in Object(arr)) { return sliceIterator(arr, i); } else { throw new TypeError('Invalid attempt to destructure non-iterable instance'); } }; })();
 
+	var _extends = Object.assign || function (target) { for (var i = 1; i < arguments.length; i++) { var source = arguments[i]; for (var key in source) { if (Object.prototype.hasOwnProperty.call(source, key)) { target[key] = source[key]; } } } return target; };
+
+	function _toConsumableArray(arr) { if (Array.isArray(arr)) { for (var i = 0, arr2 = Array(arr.length); i < arr.length; i++) arr2[i] = arr[i]; return arr2; } else { return Array.from(arr); } }
+
+	function _defineProperty(obj, key, value) { if (key in obj) { Object.defineProperty(obj, key, { value: value, enumerable: true, configurable: true, writable: true }); } else { obj[key] = value; } return obj; }
+
 	var _util = __webpack_require__(2);
 
 	var _inject = __webpack_require__(3);
+
+	// TODO(emily): Make a 'production' mode which doesn't prepend the class name
+	// here, to make the generated CSS smaller.
+	var makeClassName = function makeClassName(key, vals) {
+	    return key + '_' + (0, _util.hashObject)(vals);
+	};
+
+	// Find all of the references to descendant styles in a given definition,
+	// generates a class name for each of them based on the class name of the
+	// parent, and stores that name in a special `_names` object on the style.
+	var findAndTagDescendants = function findAndTagDescendants(styles, base) {
+	    var names = arguments.length <= 2 || arguments[2] === undefined ? {} : arguments[2];
+
+	    Object.keys(styles).forEach(function (key) {
+	        if (key[0] === ':' || key[0] === '@') {
+	            findAndTagDescendants(styles[key], base, names);
+	        } else if (key[0] === '>' && key[1] === '>') {
+	            findAndTagDescendants(styles[key], base, names);
+
+	            var _name = base + '__' + key.slice(2);
+
+	            names[key.slice(2)] = {
+	                _isPlainClassName: true,
+	                _className: _name
+	            };
+
+	            styles[key]._names = _defineProperty({}, _name, true);
+	        }
+	    });
+
+	    return names;
+	};
 
 	var StyleSheet = {
 	    create: function create(sheetDefinition) {
@@ -72,12 +110,12 @@ module.exports =
 	            var key = _ref2[0];
 	            var val = _ref2[1];
 
-	            return [key, {
-	                // TODO(emily): Make a 'production' mode which doesn't prepend
-	                // the class name here, to make the generated CSS smaller.
-	                _name: key + '_' + (0, _util.hashObject)(val),
+	            var name = makeClassName(key, val);
+
+	            return [key, _extends({
+	                _name: name,
 	                _definition: val
-	            }];
+	            }, findAndTagDescendants(val, name))];
 	        });
 	    },
 
@@ -105,6 +143,10 @@ module.exports =
 	    }
 	};
 
+	var isPlainClassName = function isPlainClassName(def) {
+	    return def._isPlainClassName;
+	};
+
 	var css = function css() {
 	    for (var _len = arguments.length, styleDefinitions = Array(_len), _key = 0; _key < _len; _key++) {
 	        styleDefinitions[_key] = arguments[_key];
@@ -121,14 +163,29 @@ module.exports =
 	        return "";
 	    }
 
-	    var className = validDefinitions.map(function (s) {
+	    // Filter out "plain class name" arguments, which just want us to add a
+	    // classname to the end result, instead of generating styles.
+	    var plainClassNames = validDefinitions.filter(isPlainClassName).map(function (def) {
+	        return def._className;
+	    });
+
+	    var otherDefinitions = validDefinitions.filter(function (def) {
+	        return !isPlainClassName(def);
+	    });
+
+	    // If there are only plain class names, just join those.
+	    if (otherDefinitions.length === 0) {
+	        return plainClassNames.join(" ");
+	    }
+
+	    var className = otherDefinitions.map(function (s) {
 	        return s._name;
 	    }).join("-o_O-");
-	    (0, _inject.injectStyleOnce)(className, '.' + className, validDefinitions.map(function (d) {
+	    (0, _inject.injectStyleOnce)(className, '.' + className, otherDefinitions.map(function (d) {
 	        return d._definition;
 	    }));
 
-	    return className;
+	    return [className].concat(_toConsumableArray(plainClassNames)).join(" ");
 	};
 
 	exports['default'] = {
@@ -179,6 +236,16 @@ module.exports =
 	};
 
 	exports.mapObj = mapObj;
+	// Removes duplicates from a list (only works for strings, non-destructive)
+	var uniquify = function uniquify(list) {
+	    var set = {};
+	    list.forEach(function (k) {
+	        return set[k] = true;
+	    });
+	    return Object.keys(set);
+	};
+
+	exports.uniquify = uniquify;
 	var UPPERCASE_RE = /([A-Z])/g;
 	var MS_RE = /^ms-/;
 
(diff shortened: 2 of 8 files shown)


README.md
```diff
@@ -173,6 +173,37 @@ const styles = StyleSheet.create({
...
+                // Styles to apply to the child when the parent is
+                // hovered over.
+                color: "red",
+                fontWeight: "bold"
+            }
+        }
+    }
+});
+```
+
```
kentcdodds: Could we add the example css output of this?

README.md
```diff
@@ -173,6 +173,37 @@ const styles = StyleSheet.create({
...
 
+# Descendant selectors
+
+Descendant rules (i.e. a rule like `.parent:hover .child { ... }` in CSS) are created by adding values to your styles indicating the styles to apply to the child. To distinguish the styles from normal CSS keys, the keys start with a `>>`, like `'>>child'`.
+
+```js
+const styles = StyleSheet.create({
+    parent: {
+        ":hover": {
+            ">>my_child": {
```
kentcdodds: personal preference: use `myChild` rather than `my_child` just because I don't like `snake_case` in my JavaScript (it's common to lint against it) and I'd prefer examples to use a style that's generally acceptable for JavaScript object property names.

README.md
```diff
@@ -173,6 +173,37 @@ const styles = StyleSheet.create({
 
 Aphrodite will ensure that the global `@font-face` rule for this font is only inserted once, no matter how many times it's referenced.
 
+# Descendant selectors
+
+Descendant rules (i.e. a rule like `.parent:hover .child { ... }` in CSS) are created by adding values to your styles indicating the styles to apply to the child. To distinguish the styles from normal CSS keys, the keys start with a `>>`, like `'>>child'`.
+
+```js
+const styles = StyleSheet.create({
```
kentcdodds: This is great that we could accomplish this without having to have a special API for child selectors :+1:

dist/aphrodite.js
```diff
@@ -60,10 +60,48 @@ module.exports =
 
 	var _slicedToArray = (function () { function sliceIterator(arr, i) { var _arr = []; var _n = true; var _d = false; var _e = undefined; try { for (var _i = arr[Symbol.iterator](), _s; !(_n = (_s = _i.next()).done); _n = true) { _arr.push(_s.value); if (i && _arr.length === i) break; } } catch (err) { _d = true; _e = err; } finally { try { if (!_n && _i['return']) _i['return'](); } finally { if (_d) throw _e; } } return _arr; } return function (arr, i) { if (Array.isArray(arr)) { return arr; } else if (Symbol.iterator in Object(arr)) { return sliceIterator(arr, i); } else { throw new TypeError('Invalid attempt to destructure non-iterable instance'); } }; })();
 
+	var _extends = Object.assign || function (target) { for (var i = 1; i < arguments.length; i++) { var source = arguments[i]; for (var key in source) { if (Object.prototype.hasOwnProperty.call(source, key)) { target[key] = source[key]; } } } return target; };
```
kentcdodds: I would expect the generated files to only be generated in a release commit. Helps keep the PRs scoped to the actual changes.

src/generate.js
```diff
@@ -2,39 +2,89 @@ import Prefixer from 'inline-style-prefixer';
...
         if (key[0] === ':') {
-            pseudoStyles[key] = merged[key];
+            pseudoStyles[key] = style[key];
         } else if (key[0] === '@') {
-            mediaQueries[key] = merged[key];
+            mediaQueries[key] = style[key];
+        } else if (key[0] === '>' && key[1] === '>') {
+            // So we don't generate weird "_names: [Object object]" styles,
+            // make a copy of the styles and get rid of the _names value.
+            const stylesWithoutNames = {
```
kentcdodds: You could avoid the use of `delete` (which has negative perf implications) by doing this:

```javascript
const {_names, ...stylesWithoutNames} = style[key]
```

[For example](http://babeljs.io/repl/#?evaluate=true&presets=es2015%2Creact%2Cstage-2&experimental=false&loose=false&spec=false&code=const%20style%20%3D%20%7B%0A%20%20bar%3A%20%7B%0A%20%20%20%20_names%3A%20'blah'%2C%0A%20%20%20%20otherProperty%3A%201%2C%0A%20%20%20%20otherProp%3A%202%2C%0A%20%20%7D%0A%7D%0A%0Aconst%20key%20%3D%20'bar'%0A%0A%2F%2F%20the%20relevant%20code%3A%0Aconst%20%7B_names%2C%20...stylesWithoutNames%7D%20%3D%20style%5Bkey%5D%0Aconsole.log(stylesWithoutNames))

:-)

kentcdodds:
Tried to review where I could. I'm afraid I still have a very limited understanding of how all this stuff works, but most stuff looks good to me :D

src/generate.js
xymostech: :O that's so much cleaner!

README.md
```diff
@@ -173,6 +173,37 @@ const styles = StyleSheet.create({
...
+    parent: {
```
jlfwong: Both in the original spec you described and in this documentation, it wasn't clear to me whether `parent` was just a style name or whether `parent` carries some special meaning. It seems to be the latter, but is worth clarifying and perhaps using a different name.

dist/aphrodite.js
jlfwong: Eugh, this should be collapsed as a generated file from setting the `.gitattributes`. I'm not sure why it's not.

src/generate.js
```diff
@@ -2,39 +2,89 @@ import Prefixer from 'inline-style-prefixer';
 
 import {
     objectToPairs, kebabifyStyleName, recursiveMerge, stringifyValue,
-    importantify
+    importantify,
 } from './util';
 
+// In a bundle of styles, find all of the different class names that a single
```
jlfwong: What's a "bundle of styles"?

README.md
```diff
@@ -173,6 +173,37 @@ const styles = StyleSheet.create({
 
 Aphrodite will ensure that the global `@font-face` rule for this font is only inserted once, no matter how many times it's referenced.
 
+# Descendant selectors
+
+Descendant rules (i.e. a rule like `.parent:hover .child { ... }` in CSS) are created by adding values to your styles indicating the styles to apply to the child. To distinguish the styles from normal CSS keys, the keys start with a `>>`, like `'>>child'`.
```
jlfwong: On initial read of this, I assumed that `child` would be a literal class name and it took me a couple read throughs to figure out that `child` would be a classname generated by Aphrodite.

It might be worth explaining here that this is really only for cases that can't be expressed directly through others mean (which, I believe, is only styling children of pseudo-selector-modified parents).

jlfwong:
Thanks for giving this so much thought in order to implement missing functionality while not compromising on the philosophy of Aphrodite, and thanks to @kentcdodds for the quick review!

I'm probably going to need to take multiple passes at this, but I'll start with the helpfully thorough PR description.

Can you start out the PR description with what this is trying to accomplish, not with what makes it difficult? If I understand correctly, this is specifically to address the fact that children of pseudo selectors are not implementable without hacks in Aphrodite, right? Even starting off by saying it fixes #10 would've been helpful (I see that it's indeed present at the bottom of the description).

```
.parent_abcdef-o_O-altparent_123456:hover .parent_abcdef__child { ... }
.parent_abcdef-o_O-altparent_123456:hover .altparent_123456__child { ... }
```

The `...` in this case will be identical, right? If I'm understanding correctly, the hashes would end up being the same (the `abcdef` vs `123456` has me a little confused). If it's true that the hashes match, it might be better to jut throw out the name wholesale and avoid the need to generate duplicate styles, especially since we were tinkering with the idea of doing this on all styles to reduce bytes sent over the wire.

src/index.js
```diff
@@ -5,14 +5,43 @@ import {
...
+    return names;
+};
+
 const StyleSheet = {
     create(sheetDefinition) {
         return mapObj(sheetDefinition, ([key, val]) => {
+            const name = makeClassName(key, val);
+
             return [key, {
-                // TODO(emily): Make a 'production' mode which doesn't prepend
```
jlfwong: This is still relevant, isn't it?

src/generate.js
```diff
@@ -2,39 +2,89 @@ import Prefixer from 'inline-style-prefixer';
...
 
+    const classNamesForDescendant = {};
+    findNamesForDescendants(merged, classNamesForDescendant);
+
+    return generateCSSInner(
+        selector, merged, stringHandlers, useImportant,
+        classNamesForDescendant);
+}
+
+const generateCSSInner = (selector, style, stringHandlers,
```
jlfwong: This is getting pretty complicated, and could also be greatly aided by a JSdoc, IMO.

tests/index_test.js
```diff
@@ -118,6 +119,44 @@ describe('css', () => {
...
+                }
+            },
+
+            blue: {
+                color: 'blue',
+            },
+        });
+
+        assert.equal(
+            css(sheet.blue, sheet.base.child),
```
jlfwong: What happens if you have `css(sheet.base.child, sheet.blue)`? I'm guessing behaviour is reasonable, but would be good to encode in a test.

montemishkin:
> Thanks for giving this so much thought in order to implement missing functionality while not compromising on the philosophy of Aphrodite, and thanks to @kentcdodds for the quick review!

> I'm probably going to need to take multiple passes at this...

Agreed!

Thoughts from my first pass:

Given your approach, and taking into account the helpful notes from @jlfwong and @kentcdodds, the implementation looks good!

I'll try to find some time ASAP to both take a deeper dive into the current implementation and also consider the overall design / approach.

natew:
Just want to say this is a really great idea, especially for the `:hover>child` which is a really common pattern I've run into.

Why not just use `>` rather than `>>`?

kentcdodds:
> Why not just use > rather than >>?

There was discussion on that in the issue. Because `>` has meaning in CSS (direct child) and this would not follow that meaning (any descendant) we had to come up with something else.

README.md
xymostech: Done!

README.md
xymostech: Oops, my python was leaking. Done!

src/generate.js
xymostech: Done!

README.md
xymostech: Done!

src/generate.js
xymostech: Done! (and in a couple other places as well)

tests/index_test.js
xymostech: done!

xymostech:
Okay, I got around to cleaning up this pull request a bit. Not much changed, but I added a bunch more docs so hopefully that should clear up some confusion.

@jlfwong, some responses to your comments:
> Can you start out the PR description with what this is trying to accomplish, not with what makes it difficult? If I understand correctly, this is specifically to address the fact that children of pseudo selectors are not implementable without hacks in Aphrodite, right? Even starting off by saying it fixes #10 would've been helpful (I see that it's indeed present at the bottom of the description).

I added a better intro to the description, hopefully that should be more helpful about what's going on. Sorry, I was approaching this with all of the context in #10!

> ```
.parent_abcdef-o_O-altparent_123456:hover .parent_abcdef__child { ... }
.parent_abcdef-o_O-altparent_123456:hover .altparent_123456__child { ... }
````
The ... in this case will be identical, right? If I'm understanding correctly, the hashes would end up being the same (the abcdef vs 123456 has me a little confused). If it's true that the hashes match, it might be better to jut throw out the name wholesale and avoid the need to generate duplicate styles, especially since we were tinkering with the idea of doing this on all styles to reduce bytes sent over the wire.

You are correct, the `...` is identical. Actually now that I think about it, maybe I should be combining those with a comma (somehow forgot that CSS supports that!) The hashes would not be the same here, since they're just using the hashes we get from hashing the `parent` and `altparent` objects.

**Comment by xymostech** [comment]
Okay, I pushed a change that uses commas to join things instead of duplicating styles, hopefully that's a little better.

I also ran a code coverage tool and found a case I hadn't tested. I'm gonna go back and add tests for the other things we aren't covering right now, maybe we could add a test that ensures aphrodite stays at 100% code coverage!

**Comment by kentcdodds** [comment]
> maybe we could add a test that ensures aphrodite stays at 100% code coverage!

You may consider adding a `check-coverage` script and add that as a git hook using [ghooks](https://npmjs.com/package/ghooks) ([like this](https://egghead.io/lessons/javascript-how-to-write-a-javascript-library-adding-code-coverage-checking?series=how-to-write-an-open-source-javascript-library), though these day's I recommend [nyc](https://npmjs.com/package/nyc)). I'd recommend also using [opt-cli](https://npmjs.com/package/opt-cli) so new contributors can get involved without being blocked by coverage.

Just some thoughts :-)

File: src/generate.js [hunk 12]
```diff
@@ -2,39 +2,177 @@ import prefixAll from 'inline-style-prefix-all';
 
 import {
     objectToPairs, kebabifyStyleName, recursiveMerge, stringifyValue,
-    importantify
+    importantify, uniquify, mapObj,
 } from './util';
 
+/**
```
jlfwong: Awww yeaaahhh JSdoc  💖

src/generate.js
```diff
@@ -2,39 +2,177 @@ import prefixAll from 'inline-style-prefix-all';
...
+ * Generate a map from descendant selector names to all of the classnames that
+ * could be used for that selector.
+ *
+ * In StyleSheet.create, we add a special `_names` object to each place where a
+ * descendant style is used containing the classname associated with that
+ * descenant. Here, we traverse the styles and make a map from the descendant
```
jlfwong: s/descenant/descendant

src/generate.js
```diff
@@ -2,39 +2,177 @@ import prefixAll from 'inline-style-prefix-all';
...
+ * selectors to all of the classnames in those `_names` objects. (Since we
+ * merge together many styles that come out of StyleSheet.create, and we store
+ * the classnames as keys on the `_names` object, the classnames that we want
+ * all end up as keys on the final merged object.)
+ *
+ * For example, given styles looking like:
+ * ```
```
jlfwong: Since this is a private API, and therefore unlikely to be nicely presented as HTML docs, I don't think we need markdown indicators. (If you do want to keep them, you're missing a closing triple backtick before the `@returns` below)

src/generate.js
```diff
@@ -2,39 +2,177 @@ import prefixAll from 'inline-style-prefix-all';
...
+ *   ":active": {
+ *     fontWeight: "bold",
+ *     ">>bar": {
+ *       _names: { "foo_bar": true },
+ *       height: 10,
+ *     }
+ *   }
+ * }, ...);
+ * ```
+ * will make 5 calls to `generateCSSRuleset`:
```
jlfwong: Super helpful example 🌟

kentcdodds:
FYI, this PR can no longer merge as it has conflicts.

src/generate.js
```diff
@@ -55,6 +193,27 @@ const runStringHandlers = (declarations, stringHandlers) => {
...
 
+/**
+ * Generate a CSS ruleset with the selector and containing the declarations.
+ *
+ * This function assumes that the given declarations don't contain any special
+ * children (such as media queries, pseudo-selectors, or descendant styles).
+ *
+ * Example:
+ * ```
+ * generateCSSRuleset(".blah", { color: "red" });
```
jlfwong: Would be good to include an example that involves a pseudo selector so show that `selector` isn't necessarily just a classname selector

src/index.js
```diff
@@ -5,14 +5,43 @@ import {
...
+            findAndTagDescendants(styles[key], base, names);
+
+            const name = `${base}__${key.slice(2)}`;
+
+            names[key.slice(2)] = {
+                _isPlainClassName: true,
+                _className: name,
+            };
+
+            styles[key]._names = { [name]: true };
```
jlfwong: I'm getting a little confused here. Why is this a map if it only ever contains one key/value pair? Am I missing something here? I see above there's a traversal in `findNamesForDescendants` throughout the whole `styles` structure to find all the ancestors of a given `>>child` selector, but I don't see how it could have more than a single name in a single location here. This suspicion is supported by the fact that only a single assignment of `_names` takes place at `StyleSheet.create` time as far as I can tell.

The example above shows:

```
+ *   ">>child": {
 + *     _names: { "parent1_child_x": true, "parent2_child_y": true },
 + *     color: "red"
 + *   },
```

How did that `_names` with two keys happen given this code?

src/index.js
```diff
@@ -49,11 +80,23 @@ const css = (...styleDefinitions) => {
         return "";
     }
 
-    const className = validDefinitions.map(s => s._name).join("-o_O-");
+    // Filter out "plain class name" arguments, which just want us to add a
+    // classname to the end result, instead of generating styles.
```
jlfwong: I think it's probably helpful to say `this is the case, e.g., for descendent selectors where the relevant CSS is generated by the call to `css(styles.foo)`, not the call to `css(styles.foo.someChild)`.

tests/generate_test.js
```diff
@@ -128,4 +128,120 @@ describe('generateCSS', () => {
...
+                color: 'green',
+                _names: {
+                    'foo_123456__blue': true,
+                },
+            },
+        }], '.foo .foo_abcdef__blue,' +
+                  '.foo .foo_123456__blue{color:green !important;}');
+    });
+
+    it('handles multiples of the same descendant name', () => {
```
jlfwong: I'm not sure what "multiples" refers to here

tests/generate_test.js
```diff
@@ -128,4 +128,120 @@ describe('generateCSS', () => {
...
+                    _names: {
+                        'foo__blue': true,
+                    },
+                },
+            },
+        }], '.foo:hover .foo__blue{color:green !important;}' +
+                  '.foo .foo__blue{color:blue !important;}');
+    });
+
+    it('handles merging and multiples for descenant styles', () => {
```
jlfwong: s/descenant/descedant

jlfwong:
Okay, after a second pass, I think I have a better understanding of what this is doing, but am still fuzzy on a few details, and am becoming a little uneasy about some behaviour (if I understand correctly how this works) that would be surprising to me.

Here are my major reservations about the API design:

## If you use a child selector without using the associated parent selector, nothing happens.

If I'm understanding the code correctly, if you have the following:

```
    foo: {
        '>>someChild': { color: 'red'; }
    }
});
```

And in your component, you have:

```
<div>
     <div className={css(styles.foo.someChild)}>Yo</div>
</div>
```

Then the div containing `Yo` will receive no styles, but WILL receive a class name. This is especially bad when combined with a silent failure from a typo, like so:

```
<div className={css(styles.fo)}>
     <div className={css(styles.foo.someChild)}>Yo</div>
</div>
```

## Because of multiple names, there are multiple ways of referencing the same child.

Having multiple ways of accomplishing the same thing generally acts as a red flag of API design to me, but perhaps it's not sensibly avoidable in this case...?

```
    foo: {
        '>>someChild': { color: 'red'; }
    },
    bar: {
        '>>someChild': { color: 'green'; }
    }
});
```

Then, if I'm understanding correctly, the following are equivalent from a user perspective and will generate the same CSS (though not HTML?):

```
<div className={css(styles.foo, styles.bar)}>
     <div className={css(styles.foo.someChild)}>
</div>
```

and

```
<div className={css(styles.foo, styles.bar)}>
     <div className={css(styles.bar.someChild)}>
</div>
```

and

```
<div className={css(styles.foo, styles.bar)}>
     <div className={css(styles.bar.someChild, styles.foo.someChild)}>
</div>
```

## The semantics of nested descendent selectors is confusing to me

...or possibly not implemented correctly? The code goes through all the trouble of identifying what the possibly parent selectors might be, but from the looks of it, it wasn't done recursively? If you have this:

```
    a1: {
        '>>b': { '>>c': { color: 'red'; } }
    },
    a2: {
        '>>b': { '>>c': { color: 'green'; } }
    }
});
```

Then what happens if you do this?

```
<div className={css(a1, a2)}>
     <div className={css(a1.b)}>
        <div className={css(a2.c})>
        </div>
    </div>
</div>
```

Overall, the semantics of any examples involving multiple instances of `>>someChild` aren't terribly clear to me (both in mergers and in pseudo selectors). Are there any valid use cases of allowing nested descendent selectors that aren't accomplishable in some other way? How much simpler does the design become if we restrict descendant selectors to only being immediate children of pseudo-selectors, and not allowing nesting of descendant selectors?

It's also not intuitive to me that `>>b` is addressed as `a1.b`, but `>>c` is not addressed as `a1.b.c`, though I guess it's okay because `a2` is not the same kind of thing as `b`?

Do others find the semantics from a user-standpoint confusing here? I'm particularly interested in @zgotsch thoughts, who may be able to draw from his experience in a real codebase that resulted in #10 being opened in the first place.

From a code complexity standpoint, this change seriously impedes my ability to reason intuitively about what the behavior of the code will be in all cases, but I didn't see any unnecessary complexity given this API design, so it may be necessary evil.

src/index.js
xymostech: More than one name could appear here when they get merged together by a call to `css()`. So if you have like
```
    a: {
        '>>child': {
            color: 'red'
        }
    },

    b: {
        '>>child': {
            color: 'blue'
        }
    }
});

css(styles.a, styles.b);
```
then during the call to `css()` when we merge the `a` styles and `b` styles, the `_names` object will end up containing the classname from `a.>>child` and `b.>>child`. Does that make sense?

jlfwong: That's how you wind up with more than one value in the array produced by `findNamesForDescendants`, but that doesn't yield multiple keys in the `_names` object, right?

xymostech:
@jlfwong Thank you for the review of this API! I agree, it's kinda confusing. I'm also super sad that this change makes the code more difficult to understand, so I'd love to figure out an API that resulted in more intuitive code. I couldn't come up with a better way to do it myself, but maybe together we can brainstorm something? Gonna respond to your different points:

> If you use a child selector without using the associated parent selector, nothing happens.

That is correct. I'm not super worried about this one though, because it's the same as how normal CSS works, no? It does make the "these descendant selectors are *_special_*" more apparent though.

> Because of multiple names, there are multiple ways of referencing the same child.

Yeah, this is fugly. You are correct, those are all style-equivalent. I guess I chose this because it seems like it would lead to the least amount of things-not-working. If you have some parent selectors and some associated child selectors, then any combination should Just Work. It seemed simpler from an external API perspective, but leads to complex code.

> The semantics of nested descendent selectors is confusing to me

They're kinda confusing to me too. It wouldn't be too much work to make `a1.b.c` work, but it's not clear that that's any better. We could just disallow nested child selectors, but that might be more confusing? I guess if we threw an error that would be okay.

wrt. your example, I have no idea what happens there. Actually maybe that doesn't work any more now that I implemented the comma stuff. Let's go check!

------

I guess overall, I really like the API for simple cases but gets much more confusing with complex cases. Maybe something similar but simplified would be great? I'm not at all attached to this implementation, it just ended up having a really clean external API (as opposed to the `cssWithDescendants` stuff, which ended up being really hard to use).

iamrane:
Why not using the more sass syntax
'& .child': {
   padding: 10
}

xymostech:
@iamrane The reasons we're not landing this is because the API is too complex/confusing. Once that gets worked out, we could figure out what syntax we would like to use.

aesopwolf:
Any thoughts on using just a space as the child selector instead of two angle brackets `>>child`?

```js
' child': {
    color: "red"
},
```

marcusnielsen:
@aesopwolf I feel white spaces are dangerous due to the low visibility. Try find a bug where you missed a white space or had the wrong number or wrong kind. It's not fun. Same issue with O/0 and other ambiguous characters.

xymostech:
Okay, I'm going to close this PR since it's stagnated a bunch and I'm not super enthusiastic that this approach is going to work. All discussion should continue in #10.

PR #71, opened 2016-05-03
Upgrade prefixer, don't handle multiple rules at once

### LINKED ISSUE #70: CSS rules with semicolons break due to importantify.
We ran into an issue here at Spring when upgrading to the newest version of Aphrodite.

We use inline data uris for image icons, with styles like:

```
cartIcon: {
  backgroundImage: 'url(data:image/svg+xml;base64,PD94...',
  ...
}
```

which produced

```
.cartIcon-somehash {
  background-image: url(data:image/svg+xml !important;base64,...
  ...
}
```

I looked at the code and noticed it would also fail on something like:

```
content: "'my content;'",
```

I have a fix, and will PR.

zgotsch:
Use the new `inline-style-prefix-all`, which gives us a nice array of rules which need to be prefixed. Only importantify one rule at a time, so we don't have to parse css in order to pick the right semicolons :)

fixes #70

### DIFF
--- a/src/generate.js
+++ b/src/generate.js
@@ -64,7 +64,7 @@ export const generateCSSRuleset = (selector, declarations, stringHandlers,
 
     const prefixedRules = flatten(
         objectToPairs(prefixedDeclarations).map(([key, value]) => {
-            if (typeof value === 'object') {
+            if (Array.isArray(value)) {
                 // inline-style-prefix-all returns an array when there should be
                 // multiple rules, we will flatten to single rules
                 return value.map(v => [key, v]);
--- a/src/generate.js
+++ b/src/generate.js
@@ -2,7 +2,7 @@ import prefixAll from 'inline-style-prefix-all';
 
 import {
     objectToPairs, kebabifyStyleName, recursiveMerge, stringifyValue,
-    importantify
+    importantify, flatten
 } from './util';
 
 export const generateCSS = (selector, styleTypes, stringHandlers,
@@ -62,7 +62,18 @@ export const generateCSSRuleset = (selector, declarations, stringHandlers,
 
     const prefixedDeclarations = prefixAll(handledDeclarations);
 
-    const rules = objectToPairs(prefixedDeclarations).map(([key, value]) => {
+    const prefixedRules = flatten(
+        objectToPairs(prefixedDeclarations).map(([key, value]) => {
+            if (typeof value === 'object') {
+                // inline-style-prefix-all returns an array when there should be
+                // multiple rules, we will flatten to single rules
+                return value.map(v => [key, v]);
+            }
+            return [[key, value]];
+        })
+    );
+
+    const rules = prefixedRules.map(([key, value]) => {
         const stringValue = stringifyValue(key, value);
         const ret = `${kebabifyStyleName(key)}:${stringValue};`;
         return useImportant === false ? ret : importantify(ret);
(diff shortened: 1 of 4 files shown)


zgotsch:
I probably need to sign the CLA now, will do that in a bit.

src/generate.js
```diff
@@ -62,7 +62,18 @@ export const generateCSSRuleset = (selector, declarations, stringHandlers,
 
     const prefixedDeclarations = prefixAll(handledDeclarations);
 
-    const rules = objectToPairs(prefixedDeclarations).map(([key, value]) => {
+    const prefixedRules = flatten(
+        objectToPairs(prefixedDeclarations).map(([key, value]) => {
+            if (typeof value === 'object') {
```
xymostech: Maybe use `Array.isArray`?

src/util.js
```diff
@@ -180,14 +180,13 @@ function murmurhash2_32_gc(str) {
...
+const IMPORTANT_RE = /^([^:]+:.*?)( !important)?;$/;
+
+// Given a single style rule string like "a: b;", adds !important to generate
+// "a: b !important;".
+export const importantify = (string) =>
+    string.replace(
+        IMPORTANT_RE,
+        (_, base, important) => base + " !important;");
+
+export const flatten = (list) => list.reduce((memo, x) => memo.concat(x), []);
```
xymostech: Could you add a comment that shows the types of this function (like we do for `pairsToObject` above)?

xymostech:
This looks awesome!!

Could you add a test to see how this interacts with our `stringHandlers` stuff? For instance, our `fontFamily` string handler also accepts arrays as a valid value, and I want to make sure that this doesn't clobber that (we unfortunately don't have tests for `stringHandlers` yet, maybe that would answer my question).

I guess something that looks like
```
assertCSS('.foo', [{
    test: ['a', 'b'],
}], ".foo { test: a b !important; }", { test: (val) => val.join(" ") });
```
(I'm worried it would turn into `".foo { test: a !important; test: b !important; }"`, which isn't what we want)

src/generate.js
zgotsch: Yeah I thought of this but I don't know what your polyfilling setup is. If it's safe, it's way better.

zgotsch:
If I understand correctly, there is already a test for this:

```
    it('supports custom string handlers', () => {
        assertCSS('.foo', [{
            fontFamily: ["Helvetica", "sans-serif"]
        }], '.foo{font-family:Helvetica, sans-serif !important;}', {
            fontFamily: (val) => val.join(", "),
        });
    });
```

src/util.js
zgotsch: Yep

xymostech:
Oh yep, you're right! Sweet. Did you sign the CLA? (we should really hook our CLA bot up to this repo...)

zgotsch:
I did sign the CLA!

jlfwong:
I'm confused -- didn't we only need `importantify` in the first place because we weren't getting an array of values? Why do we still need the regex? Just to avoid double important'ing things?

zgotsch:
Yeah, that's why I kept using it. There should probably be a test for that
so it doesn't get broken in future.

zgotsch:
Oof, seems like prefixes are not ordered consistently across platforms. I'm going to sort them...

zgotsch:
Is this safe? In the case of something like `-webkit-box` and `-webkit-flex`, the older prefix should probably go first. It's not trivial to know the rules for ordering, though. This should also probably be the responsibility of `inline-prefix-all`, not `aphrodite`.

kentcdodds:
> This should also probably be the responsibility of inline-prefix-all, not aphrodite.

Yes

xymostech:
@zgotsch Okay, I'm gonna merge this!

PR #79, opened 2016-06-01
Add note about async buffer flushing.

pvolok:


kentcdodds:
:+1:

jlfwong:
Thanks for taking this on, @pvolok! I think this would be very well served by a simple concrete code sample.

pvolok:
Done.

jlfwong:
Great! Thanks for this :)

PR #82, opened 2016-06-05
Check for already-injected @font-face rules by src - Fixes #80

### LINKED ISSUE #80: @font-face multiple font styles
I'm trying to figure out how to include 2 font styles for a given font family.

I tried the following, but the `italicFont` *@*font-face isn't being added to the stylesheet.

```js
const font = {
  fontStyle: "normal",
  fontWeight: "normal",
  fontFamily: "My Font",
  src: `url(${myFontWoff}) format("woff"),
        url(${myFontTtf}) format("truetype")`
};

const italicFont = {
  fontStyle: "italic",
  fontWeight: "normal",
  fontFamily: "My Font",
  src: `url(${myItalicFontWoff}) format("woff"),
        url(${myItalicFontTtf}) format("truetype")`
};

fontFamily: [font, italicFont, "serif"]
```

Am I missing something or does this not yet work in Aphrodite?

jdan:
Hey @xymostech && @jlfwong,

This changeset modifies the call to `injectStyleOnce` to use `src` as the key, rather than `fontName` [which does not need to be unique](http://fonts.googleapis.com/css?family=Inconsolata:400,700).

I added a quick test as well, let me know if you'd like it to be more thorough. You can also [see the fix in action on my cloned Aphrodite output](https://output.jsbin.com/ticukov?code=%7B%0A%20%20%22test%22%3A%20%7B%0A%20%20%20%20fontFamily%3A%20%5B%7B%0A%20%20%20%20%20%20fontStyle%3A%20%22normal%22%2C%0A%20%20%20%20%20%20fontWeight%3A%20%22normal%22%2C%0A%20%20%20%20%20%20fontFamily%3A%20%22My%20Font%22%2C%0A%20%20%20%20%20%20src%3A%20%27url%28blah%29%20format%28%22woff%22%29%2C%20url%28blah%29%20format%28%22truetype%22%29%27%0A%20%20%20%20%7D%2C%20%7B%0A%20%20%20%20%20%20fontStyle%3A%20%22italic%22%2C%0A%20%20%20%20%20%20fontWeight%3A%20%22normal%22%2C%0A%20%20%20%20%20%20fontFamily%3A%20%22My%20Font%22%2C%0A%20%20%20%20%20%20src%3A%20%27url%28blahitalic%29%20format%28%22woff%22%29%2C%20url%28blahitalic%29%20format%28%22truetype%22%29%27%0A%20%20%20%20%7D%5D%0A%20%20%7D%0A%7D).

Input:
```js
{
  "test": {
    fontFamily: [{
      fontStyle: "normal",
      fontWeight: "normal",
      fontFamily: "My Font",
      src: 'url(blah) format("woff"), url(blah) format("truetype")'
    }, {
      fontStyle: "italic",
      fontWeight: "normal",
      fontFamily: "My Font",
      src: 'url(blahitalic) format("woff"), url(blahitalic) format("truetype")'
    }]
  }
}
```

Current:
```css
@font-face {
  font-style:normal;font-weight:normal;font-family:My Font;src:url(blah) format("woff"), url(blah) format("truetype");
}
.test_1pqrvdz {
  font-family:"My Font","My Font" !important;}
```

With fix (notice how "My Font" is uniquified in `font-family`):
```css
@font-face {
  font-style:normal;font-weight:normal;font-family:My Font;src:url(blah) format("woff"), url(blah) format("truetype");
}
@font-face {
  font-style:italic;font-weight:normal;font-family:My Font;src:url(blahitalic) format("woff"), url(blahitalic) format("truetype");
}
.test_1pqrvdz {
  font-family:"My Font" !important;}
```

Maybe we want to think about only adding `@font-face` if it actually applies to the `font-family` rule? (i.e., only adding the second `@font-family` if we use `font-style: italic`?) dunno


src/util.js
```diff
@@ -43,6 +43,10 @@ export const recursiveMerge = (a, b) => {
     return ret;
 };
 
+export const uniquify = (values) => {
+    return [...new Set(values)];
```
xymostech: I'm not sure we guarantee that people have a `Set` polyfill, so I don't think this will work?

src/inject.js
```diff
@@ -48,9 +48,9 @@ const stringHandlers = {
     // an array of objects and strings.
     fontFamily: function fontFamily(val) {
         if (Array.isArray(val)) {
-            return val.map(fontFamily).join(",");
+            return uniquify(val.map(fontFamily)).join(",");
```
xymostech: I don't think it's terribly important to uniquify these font names. It'll still work without this, correct?

xymostech:
This fixes #80, correct?

src/inject.js
jdan: Correct, and I'm happy to revert those last 2 commits which add this behavior if we don't want it.

jdan:
Yep fixes #80! It's in the commit message so it'll auto-close if merged.

src/util.js
jdan: Good catch! Looks like our build script doesn't polyfill it.

jlfwong: I'm guessing this was already the plan, but I'd rather just not use `Set` than change this to include the polyfill (in the interest of keeping byte size down).

jdan: Yeah we're on the same page. Since outputting unique `font-family`s isn't important we can skip all of this.

jdan:
Kk after some revert-revert-reverts I removed the unique font style stuff. This is good for a re-review.

jlfwong:
One line change accompanied by lots of tests -- this is my kinda change 😄 

LGTM, but I only skimmed -- will leave to @xymostech for sign-off.

smdern:
Any update on this?

xymostech:
Looks like I totally dropped the ball here! Looks like this still merges, so I'm gonna squash+merge. Thanks for the reminder @smdern

jlfwong:
Should this be a patch or a minor release?

Issue #76, opened 2016-05-20
Styles are injected after componentDidUpdate.


pvolok:
I'm trying to use aphrodite in my project. Everything works great so far except one scenario. I have a Tooltip component that measures children dimensions in `componentDidMount` and `componentDidUpdate`. But actual css gets injected after these functions were called. What's worse is that sometimes styles might be already injected, so dom element have unpredictable styling in above functions.

I created an issue on react to discuss how react might help with this problem: https://github.com/facebook/react/pull/6816

xymostech:
Thanks for reporting this! At KA, we do a `setTimeout(function() { /* measuring in here */ }, 0);` in a couple of places to solve this exact problem, but you're right that it isn't super obvious that this is needed, and is confusing at best. @matchu is the one who ran into this, so probably has some valuable feedback.

We were talking about maybe adding a function like `waitForStyles` or something that you could pass a callback, and it would run the callback once all the styles had been injected (or immediately if there were no styles queued to be injected). Does that sort of interface sound reasonable to you?

kentcdodds:
Until that can be addressed, should this be added to [the Caveats section](https://github.com/Khan/aphrodite#caveats) of the README?

matchu:
(My feedback is that, yes, we should do this! :D Even if our first draft is
super naive, it'll help clarify that this is an intended part of the API.)

pvolok:
In my opinion `waitForStyles` should be synchronous (will styles be applied synchronously after `style` tag injection?). It will make things simpler. What do you think?

I'd be happy to send a PR later today.

matchu:
The only way for it to be synchronous is for the `waitForStyles` call to immediately flush the buffer, right? That seems bad for performance—it's the same reason that we even _have_ a buffer and don't just flush on every single `css` call.

Even if the performance difference turned out to be small, I'd prefer the async API. The complexity difference seems almost nonexistent to me, and async gives us more options if we decide to change the implementation in the future.

pvolok:
I checked if injected styles are applied immediately. And the answer is no, unfortunately. So probably we can't even implement synchronous `waitForStyle` with current approach.

My idea of flushing styles is to follow react's process: call all `render` functions, then flush to the dom, and call all `didComponentUpdate`s. If we know we need styles in `didComponentUpdate`, we can flush all `css` calls that are already batched. So aphrodite will flush as many times as react flushes (or less).

matchu:
[...]

PR #83, opened 2016-06-06
Try insertRule

pvolok:
This PR is a demonstration of using `styleTag.sheet.insertRule` instead of `styleTag.appendChild` (#76). The code is probably bad, I just made a minimum amount of changes to show the `insertRule` approach. It's not intended to be merge, just to discuss :)

This way styles are applied immediately in the `css` function. So in `componentDidMount` and `componentDidUpdate` all styles are already applied.

Also this approach scales better as the app continues to work, because browser doesn't have to reparse all css.

Problems:
- `insertRule('::-moz-focus-inner {}', 0)` throws in chrome. If we wrap it with try-catch, we have about 25% performance penalty.
- This API doesn't exists on IE8. Does aphrodite support it?

### DIFF
--- a/src/generate.js
+++ b/src/generate.js
@@ -23,20 +23,20 @@ export const generateCSS = (selector, styleTypes, stringHandlers,
         }
     });
 
-    return (
+    return [
         generateCSSRuleset(selector, declarations, stringHandlers,
-            useImportant) +
-        Object.keys(pseudoStyles).map(pseudoSelector => {
+            useImportant),
+        ...Object.keys(pseudoStyles).map(pseudoSelector => {
             return generateCSSRuleset(selector + pseudoSelector,
                                       pseudoStyles[pseudoSelector],
                                       stringHandlers, useImportant);
-        }).join("") +
-        Object.keys(mediaQueries).map(mediaQuery => {
+        }),
+        ...Object.keys(mediaQueries).map(mediaQuery => {
             const ruleset = generateCSS(selector, [mediaQueries[mediaQuery]],
                 stringHandlers, useImportant);
-            return `${mediaQuery}{${ruleset}}`;
-        }).join("")
-    );
+            return `${mediaQuery}{${ruleset.join('')}}`;
+        }),
+    ];
 };
 
 const runStringHandlers = (declarations, stringHandlers) => {
--- a/src/inject.js
+++ b/src/inject.js
@@ -32,11 +32,13 @@ const injectStyleTag = (cssContents) => {
         }
     }
 
-    if (styleTag.styleSheet) {
-        styleTag.styleSheet.cssText += cssContents;
-    } else {
-        styleTag.appendChild(document.createTextNode(cssContents));
-    }
+    return styleTag;
+
+    // if (styleTag.styleSheet) {
+    //     styleTag.styleSheet.cssText += cssContents;
+    // } else {
+    //     styleTag.appendChild(document.createTextNode(cssContents));
+    // }
 };
 
 // Custom handlers for stringifying CSS values that have side effects
@@ -96,7 +98,7 @@ const stringHandlers = {
         });
         finalVal += '}';
 
-        injectGeneratedCSSOnce(name, finalVal);
+        injectGeneratedCSSOnce(name, [finalVal]);
 
         return name;
     },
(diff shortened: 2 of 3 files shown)


src/inject.js
```diff
@@ -105,6 +107,7 @@ const stringHandlers = {
 // This is a map from Aphrodite's generated class names to `true` (acting as a
 // set of class names)
 let alreadyInjected = {};
+global.inj = alreadyInjected;
```
kentcdodds: This wont work in the browser right?

pvolok: It will, if built with webpack. But this shouldn't have been committed :)

xymostech:
I really like this! Might as well use browser APIs to make things better when we can!

src/generate.js
```diff
@@ -23,20 +23,20 @@ export const generateCSS = (selector, styleTypes, stringHandlers,
         }
     });
 
-    return (
+    return [
         generateCSSRuleset(selector, declarations, stringHandlers,
-            useImportant) +
-        Object.keys(pseudoStyles).map(pseudoSelector => {
+            useImportant),
+        ...Object.keys(pseudoStyles).map(pseudoSelector => {
```
xymostech: does this compile to something more complex than `.concat`?

src/inject.js
```diff
@@ -32,11 +32,13 @@ const injectStyleTag = (cssContents) => {
         }
     }
 
-    if (styleTag.styleSheet) {
-        styleTag.styleSheet.cssText += cssContents;
-    } else {
-        styleTag.appendChild(document.createTextNode(cssContents));
-    }
+    return styleTag;
```
xymostech: Based on what we have here, it seems like this function should take an array of styles and should either loop over them and call `styleTag.sheet.insert`, or `.join("")` them and fallback to the current code if `styleTag.sheet` doesn't exist.

src/generate.js
```diff
@@ -23,20 +23,20 @@ export const generateCSS = (selector, styleTypes, stringHandlers,
...
             return generateCSSRuleset(selector + pseudoSelector,
                                       pseudoStyles[pseudoSelector],
                                       stringHandlers, useImportant);
-        }).join("") +
-        Object.keys(mediaQueries).map(mediaQuery => {
+        }),
+        ...Object.keys(mediaQueries).map(mediaQuery => {
```
xymostech: How does `styleSheet.insertRule` work with media queries? Here, we sometimes have large blocks of media queries with muliple rules inside them, like
```
@media screen {
    .a { color: red; }
    .b { color: blue; }
}
```
Are those added appropriately?

src/generate.js
kentcdodds: Personally, I feel like this code could be refactored a bit to make it easier to read. Each item in this array should probably be assigned to a variable with a name that describes its purpose and then those could be put into the array as needed...

jlfwong:
Thanks for the PR! Would you mind updating the description of the PR (and possibly the commit message) to make it clearer what the motivation for this is and how it accomplishes it?

Many of these questions are probably answered through #76, but as much as possible I like to avoid needing to jump around to answer the question "what does this do?". It also allows an eventual merge commit to be meaningful outside the context of GitHub (e.g. when reading through git logs in a terminal).

src/generate.js
pvolok: Yes. It adds `CSSMediaRule` with multiple `CSSStyleRule` rules in it.

rainydio:
What is the reason to use prefixAll from 'inline-style-prefixer/static' instead of inserting vendor specific rules based on userAgent?

* This should solve `-moz-` problems
* Reduce CSS size
* Unclutter developer tools

I understand that there is a server rendering case. But it might be more beneficial to build user-agent specific pages.

rainydio:
I assume that's because [browser detection](bowser) takes up a lot of space. If `inline-style-prefixer` drops it, and replaces [conditions](plugins) with something smaller. They look quite small. 

```js
browser === 'firefox' && version < 15 ||
browser === 'chrome' && version < 25 ||
browser === 'safari' && version < 6.1 ||
browser === 'ios_saf' && version < 7
```

[bowser]: https://github.com/ded/bowser/blob/master/src/useragents.js
[plugins]: https://github.com/rofrischmann/inline-style-prefixer/tree/master/modules/plugins

kentcdodds:
Those URLs you've specified are broken links.

jlfwong:
You're right that a benefit of using `prefixAll` makes the code size smaller, but it also facilitates better caching for server-side rendering. Otherwise, you have to make the user agent string part of the cache key. It's also just awkward in general for server-side rendering because you'd have to figure out some way of passing along the user agent string in a way that doesn't break when multiple requests from different user agents are interleaved.

PR #87, opened 2016-06-07
Document generateCSS and generateCSSRuleset methods

jlfwong:
I'm reluctant to let #61 land because of the complexity it introduces, but would
like these docs to land (slightly modified from #61), and would prefer to not
make them landing dependent on the complexity of #61.


xymostech:
LGTM! Yay docs!

PR #90, opened 2016-06-14
Add StyleSheetTestUtils module

wyattdanger:
Adds `StyleSheetTestUtils` module with `startBufferingForTests` method which enables
components using Aphrodite to render in a DOMless environment without throwing an exception.

For our use case, this allows us to use [Enzyme](https://github.com/airbnb/enzyme) to shallow render our Aphrodite-powered components in Mocha tests by running `StyleSheetTestUtils.startBufferingForTests()` in a global `beforeEach` hook so that our tests "just work" without needing to explicitly use the `renderStatic` method in every individual test.

Tests pass and maintain 100% coverage threshold.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -39,6 +39,13 @@ const StyleSheetServer = {
     },
 };
 
+const StyleSheetTestUtils = {
+  startBufferingForTests() {
+    reset();
+    startBuffering();
+  },
+};
+
 const css = (...styleDefinitions) => {
     // Filter out falsy values from the input, to allow for
     // `css(a, test && c)`
@@ -59,5 +66,6 @@ const css = (...styleDefinitions) => {
 export default {
     StyleSheet,
     StyleSheetServer,
+    StyleSheetTestUtils,
     css,
 };
--- a/tests/index_test.js
+++ b/tests/index_test.js
@@ -2,7 +2,7 @@ import asap from 'asap';
 import {assert} from 'chai';
 import jsdom from 'jsdom';
 
-import { StyleSheet, StyleSheetServer, css } from '../src/index.js';
+import { StyleSheet, StyleSheetServer, StyleSheetTestUtils, css } from '../src/index.js';
 import { reset } from '../src/inject.js';
 
 describe('css', () => {
@@ -342,5 +342,21 @@ describe('StyleSheetServer.renderStatic', () => {
 
         const newRet = StyleSheetServer.renderStatic(emptyRender);
         assert.equal(newRet.css.content, "");
+    });
+});
+
+describe('StyleSheetTestUtils.startBufferingForTests', () => {
+    const sheet = StyleSheet.create({
+        red: {
+            color: 'red',
+        },
+        blue: {
+            color: 'blue'
+        },
+    });
+
+    it('starts buffering', () => {
+        StyleSheetTestUtils.startBufferingForTests();
+        assert.doesNotThrow(() => css(sheet.red, sheet.blue));
     });
 });


kentcdodds:
Very interesting. I don't see a reason to not include this personally.

xymostech:
Hi @wyattdanger! Is there a reason why you can't just use `StyleSheetServer.renderStatic` in your tests? I don't know how enzyme works but it seems like you could do something like

```
describe('<Foo />', () => {
  it('renders three `.foo-bar`s', () => {
    const {html: wrapper, css} = StyleSheetServer.renderStatic(() => render(<Foo />));
    expect(wrapper.find('.foo-bar').length).to.equal(3);
  });
});
```

lelandrichardson:
For additional context, this allows aphrodite to be conveniently used in a mocha context, by doing:

```js
beforeEach(() => StyleSheetServer.start());
afterEach(() => StyleSheetServer.stop());
```

This helps Airbnb by not needing to wrap every test or enzyme callsite in a `StyleSheetServer.renderStatic(...)`. Additionally, this allows people to use aphrodite in an asynchronous manner... ie, something like:

```js
StyleSheetServer.start();
AsyncRenderer.renderAsync()
  .then(html => ({ html, css: StyleSheetServer.stop() }))
  .then(({html, css}) => {
    // do something with result
  });
```

@xymostech's example would technically work, but it is difficult to require every engineer to know how to do this, and is creating some friction to the adoption of aphrodite at airbnb. By using `.start()` and `.stop()` it allows us to specify a global `beforeEach()` in our test suite so that aphrodite "just works" (which is awesome!).

We are adopting aphrodite incrementally, so right now there may be a component that doesn't use aphrodite at all and has an extensive test suite written for it. Someone might come in and refactor the component to use aphrodite, which will result in all of the tests immediately failing. `renderStatic` can indeed be used to provide a workaround, but this ends up with significant awkwardness (for instance, we *aren't* rendering the component statically in most cases, so the naming is confusing).

(Side note: I would also be happy to write a guide in your documentation or something with information on how to easily test react components that use aphrodite with enzyme.)

wyattdanger:
@xymostech I believe @lelandrichardson touched on the main reason why I'd like to avoid `StyleSheetServer.renderStatic` in my tests, which is that it'd be great for Mocha tests with Aphrodite to "just work" without explicit configuration in each and every test.

xymostech:
@lelandrichardson @wyattdanger Ah, I now see what you're looking for! Hmm, that does seem to be annoying that tests randomly break. I'm mostly hesitant about exposing these more "internal" parts because I don't know what our interfaces are going to look like in the future.

Is there a reason that you want `StyleSheetServer.stop` to output the generated styles? Are you inspecting them or was it just a result of how you split up the `renderStatic` function? If not, I think I would suggest exposing a `StyleSheetServer.stopBuffering` method or something that just completely disables style generation for a bit.

For your case, you would put `StyleSheetServer.stopBuffering()` in your `beforeEach` context, and if you ever do want to inspect the generated styles you can wrap those specific parts in `StyleSheetServer.renderStatic()`? Does that sound reasonable?

lelandrichardson:
@xymostech At this point we aren't using the generated output in any way... I mostly thought that adding a `stop()` method would lend some completeness to the change.

I'm not sure if I understand your recommendation though. It seems to me what we really need/want is a `StyleSheetServer.startBuffering()` rather than a `StyleSheetServer.stopBuffering()`, no?

wyattdanger:
@xymostech yes that sounds reasonable. We already have a method in the scope of that file named `startBuffering`, though. We could leave it at `start`, and remove the `stop` method?

wyattdanger:
@xymostech I removed the `stop` method and renamed `start` to `startBuffering`. Is this what you had in mind? This approach still fulfills our test needs.

jlfwong:
Sorry, I'm still not quite following the motivation.

I *think* what we're saying here is that you need this change in order to prevent `css` calls from trying to interact with the DOM when they want to do the injection, which is causing your tests to fail in cases where you're trying to run the tests in a DOM-less environment. Is that right?

If that is right, I support @xymostech suggestion to have a method that entirely disables style generation, and make it clear that the purpose of that function is for testing. Unless I'm misunderstanding, that method would never be useful in dev/prod -- only in tests, right?

I'd be fine with making Aphrodite have  "test mode" that could be flipped on and off. It would presumably be useful to e.g. make the classname generation not contain hashes of the contents in this mode too.

As it stands, I don't support merging the current diff, since having a `startBuffering` method without a `stopBuffering` method is super confusing, and this code would presumably only ever be used in tests.

At the very least, it should be renamed `stopBufferingForTests` or something that makes it clearer what the purpose is. Or perhaps `disableBufferingForTests`.

jlfwong:
Actually, instead of having `stopBufferingForTests`, it would, IMO, be even better to make an object that just deals with this. A `StyleSheetTestUtils` that has these utilities on it. Putting it on `StyleSheetServer` feels awkward to me because this change doesn't really have anything to do with doing things server-side.

I'll see if I can get a PR up to do something reasonable here later today, assuming I'm understanding your needs here properly.

kentcdodds:
I like the idea of test utils module. This would be great.

wyattdanger:
I also like the idea of a test utils module, but don't we actually want `startBufferingForTests`?

jlfwong:
@wyattdanger Thanks for the really quick update! I wrote up #91 to accomplish this too, along with a bit more thorough documentation.

If #91 accomplishes the goals as I understand them, I'll merge that and publish a new version of aphrodite.

If you'd like to have commit credit, I'm also happy to let you take the changes presented there and amend this pull request (particularly the documentation and the test using `asap(done)`), then close #91 and merge this, providing you're okay with also signing the CLA: https://docs.google.com/forms/d/1QIZDwzdLrqyqGkqzVcMHEFSaApt_IZkGmgSnRkArvzI/viewform

Thanks for the report, and for being so responsive in this PR!

wyattdanger:
@jlfwong thanks for driving this to completion! I'm not worried about commit credit. I'll close this in favor of https://github.com/Khan/aphrodite/pull/91

PR #91, opened 2016-06-16
Introduce StyleSheetTestUtils to make writing tests for Aphrodite easier

jlfwong:
Imagine you're trying to introduce Aphrodite gradually into an existing codebase
that already has extensive test coverage for rendering components.

If those tests are run in node without any fake DOM being constructed, your
previously working tests will now fail with the following error:

     Error: Cannot automatically buffer without a document

The introduction of StyleSheetTestUtils makes it easy to suppress such problems
by preventing Aphrodite from trying to interact with the DOM at all.

The new API would be usable in your tests like so (using Mocha syntax for
demonstration purposes):

    import { StyleSheetTestUtils } from 'aphrodite';

    beforeEach(() => {
        StyleSheetTestUtils.suppressStyleInjection();
    });

    afterEach(() => {
        StyleSheetTestUtils.clearBufferAndResumeStyleInjection();
    });

I intentionally did not expose an API returning CSS content here to allow us
flexibility to change the CSS formatting or class name generation without
considering it a breaking API change.

Meant to solve the problem presented in #90

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -22,6 +22,9 @@ const StyleSheet = {
     },
 };
 
+/**
+ * Utilities for using Aphrodite server-side.
+ */
 const StyleSheetServer = {
     renderStatic(renderFunc) {
         reset();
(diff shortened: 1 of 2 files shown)


wyattdanger:
LGTM! Thanks @jlfwong

src/index.js
```diff
@@ -59,5 +92,6 @@ const css = (...styleDefinitions) => {
 export default {
```
kentcdodds: I just noticed this. I'm afraid to say that this is using ES6 modules incorrectly and it'll bite us when this project is upgraded to Babel 6. Learn more about my painful learning experience here: http://kcd.im/tears

jlfwong: Thanks for the headsup. It's been discussed ab it before, and if/when we do decide to upgrade to Babel 6, we have this PR ready: https://github.com/Khan/aphrodite/pull/27 (which also removes this `default`).

If removing that `default` is a non-breaking change, I'm also happy to take a PR with just that.

jlfwong:
@wyattdanger Published as 0.4.0 on npm. If you'd like to add docs to the README to reflect this usage or testing methodologies in general, would be happy to take a PR in that vein!

PR #92, opened 2016-06-16
Add note about unicode characters in `content`

jaredly:
This had me stuck for a while... figured I'd leave a note for the next person

### DIFF
--- a/README.md
+++ b/README.md
@@ -218,7 +218,8 @@ const styles = StyleSheet.create({
   },
   small: {
       ':before': {
-        content: "'Aphrodite'",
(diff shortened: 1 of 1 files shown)


jlfwong:
Making the example use font-awesome is pretty weird...

Why not 'Aphrodite 💖' as the example instead? :D

spicyj:
Woah weird, can you not put unicode characters in between quotes in CSS?

jaredly:
With a heart you would just put `content: '"💖"'`, because it's renderable by a normal font. Only with an icon-font like font-awesome would you use the `\f1db` format

spicyj:
Oh okay. But "\uF1DB" in the JS string should work right?

Issue #96, opened 2016-06-30
Transition property compiles to broken CSS with prefixing


gscottolson:
```es
const styles = StyleSheet.create({
  transition: 'border-color 300ms ease-in-out',
})
```
produces:
```css
  transition: b-ms-order-color 300ms ease-in-out,b-webkit-order-color 300ms ease-in-out,border-color 300ms ease-in-out !important;
  -webkit-transition: b-webkit-order-color 300ms ease-in-out,border-color 300ms ease-in-out !important;
```

Notice that `border-color` has been mangled as `b-ms-order-color` and the like. I am also seeing `webkit-transition` styles with duplicate entries (and…mangled).

xymostech:
Freaky! This looks like a bug in our auto-prefixer though, https://github.com/rofrischmann/inline-style-prefix-all I'll go report a bug there. :)

xymostech:
Looks like there's already a bug report: https://github.com/rofrischmann/inline-style-prefix-all/issues/15

ctrlplusb:
It's been fixed. :)

jlfwong:
Would be great to have a test for this, along with a fix by upgrading `inline-style-prefix-all`! Would make a good first PR ;)

gscottolson:
@jlfwong Thanks for the suggestion. PR is up.

jlfwong:
Should be fixed in 0.4.1

PR #99, opened 2016-07-06
Update inline-style-prefixer, add transition: border-color test

gscottolson:
Fixes https://github.com/Khan/aphrodite/issues/96

### DIFF
--- a/package.json
+++ b/package.json
@@ -51,6 +51,6 @@
   },
   "dependencies": {
     "asap": "^2.0.3",
-    "inline-style-prefix-all": "^2.0.0"
+    "inline-style-prefixer": "^2.0.0"
   }
 }
--- a/src/generate.js
+++ b/src/generate.js
@@ -1,4 +1,4 @@
-import prefixAll from 'inline-style-prefix-all';
+import prefixAll from 'inline-style-prefixer/static';
 
 import {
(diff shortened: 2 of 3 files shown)


gscottolson:
Do I need to include the compiled files in the `/dist` folder, too?

kentcdodds:
< insert my regular (annoying) reminder about how it'd be handy to [not commit generated files](http://kcd.im/generated) to source control >

@gscottolson, no, I believe that the project maintainers generate that themselves for releases only.

jlfwong:
@gscottolson Woohoo! Thanks for the PR. I trust the test fails when using the old version of the autoprefixer?

Yeah, it's fine -- no need to commit the generated files.

Can you sign the Khan Academy CLA if you haven't already, please? https://docs.google.com/forms/d/1QIZDwzdLrqyqGkqzVcMHEFSaApt_IZkGmgSnRkArvzI/viewform

@kentcdodds It's for interop reasons with Khan Academy's codebase, and would be sort of a PITA to fix. I also don't work there any more, so it's not really something in my control any more either :)

gscottolson:
@jlfwong Signed the CLA. And yes, the old version fails with this test, as described in https://github.com/Khan/aphrodite/issues/96.

jlfwong:
@gscottolson Thanks!

PR #104, opened 2016-07-20
Allow disabling of !important appending via aphrodite/no-important

### LINKED ISSUE #25: use of !important on all generated styles

jlfwong:
This is a non-breaking change to make it *possible* to use Aphrodite without it automatically appending `!important` to your styles, but does not make it the *default*.

Fixes #25.

Also see discussion in #102, #41.


jlfwong:
I made it `aphrodite/lib/without-bang-important` because I couldn't figure out how to make it `aphrodite/without-bang-important`, but I'm sure it's possible, so would appreciate some guidance.

Please refrain in this PR from discussing whether or not you think `!important` should be default behaviour or not, unless you are opposed to having this as an incremental step at all. This does not preclude the possibility of changing the default behaviour in the future.

spicyj:
You need a file in the root of the built package, which with the current setup probably means the root of the repo. It can be as simple as

```
module.exports = require('./lib/without-bang-important');
```

src/without-bang-important.js
```diff
@@ -0,0 +1,23 @@
...
+
+import {
+    StyleSheet,
+    StyleSheetServer,
+    StyleSheetTestUtils,
+} from './index.js'
+
+const css = (...styleDefinitions) => {
+    // Do not append !important to style definitions
+    return injectAndGetClassName(false, styleDefinitions);
```
kentcdodds: To make it so you don't need the comment above, I would recommend giving that boolean a name:

```javascript
const useImportant = false;
return injectAndGetClassName(useImportant, styleDefinitions);
```

src/index.js
```diff
@@ -73,20 +73,8 @@ const StyleSheetTestUtils = {
...
-        return "";
-    }
-
-    const className = validDefinitions.map(s => s._name).join("-o_O-");
-    injectStyleOnce(className, `.${className}`,
-        validDefinitions.map(d => d._definition));
-
-    return className;
+    // By default, append !important to all style definitions
+    return injectAndGetClassName(true, styleDefinitions);
```
kentcdodds: Same here as commented in the other useage

kentcdodds:
This looks great to me

src/without-bang-important.js
jlfwong: I see your intent, but I think the comment is more descriptive than that variable name, and it ends up being the same # of lines. If this thing had many arguments of ambiguous meaning, I'd be more inclined to the go the variable naming route (possibly still keeping the comments above the variable declarations.

src/inject.js
```diff
@@ -182,3 +182,29 @@ export const addRenderedClassNames = (classNames) => {
         alreadyInjected[className] = true;
     });
 };
+
+/**
+ * Inject styles associated with the passed style definition objects, and return
+ * an associated CSS class name.
+ *
+ * @param {boolean} useImportant If true, will
```
jlfwong: Whoops, forgot the rest of this sentence. Will fix.

src/without-bang-important.js
kentcdodds: I just prefer to avoid comments where I can because they can go out of date. Also, unless you know the function signature, you don't know that this comment is referring to the `false` argument... It almost looks like a `todo` comment that's missing the `TODO` pragma.

jlfwong: In personal experience, comments only tend to get out of date when they're far removed from the thing they apply to spacially (e.g. a comment discussion caveats of function use). I've rarely seen comments a single line above what they apply to get out of date when you have a sane code reviewing process in place.

I do see your point about the non-obvious connection between the comment and the `false` here, so I'm happy to take the hybrid approach and include the variable too.

spicyj:
nit: a vote from me for `aphrodite/no-important` for brevity; I think it's about as clear.

src/index.js
```diff
@@ -73,20 +73,8 @@ const StyleSheetTestUtils = {
...
-    if (validDefinitions.length === 0) {
-        return "";
-    }
-
-    const className = validDefinitions.map(s => s._name).join("-o_O-");
-    injectStyleOnce(className, `.${className}`,
-        validDefinitions.map(d => d._definition));
-
-    return className;
+    const useImportant = true;   // Append !important to all style definitions
```
kentcdodds: yes, this is much more clear :+1:

JedWatson:
> a vote from me for aphrodite/no-important for brevity; I think it's about as clear.

+1 from me too. I've never heard `!` called "bang"s before now, they're just "exclamation marks". I read `!important` as "exclamation-important", so I suspect "no important" is clearer internationally.

Out of interest, I [looked it up](https://en.wikipedia.org/wiki/Exclamation_mark)

> This punctuation mark is called, in the printing world, a screamer, a gasper, a slammer, or a startler. In hacker culture, the exclamation mark is called "bang", "shriek", or, in the British slang known as Commonwealth Hackish, "pling".

jlfwong:
Switched to `no-important`, thanks for the feedback, @spicyj and @JedWatson!

xymostech:
LGTM!

PR #105, opened 2016-07-21
Fix recursive merge when `a` has a null property

samiskin:
This fixes the case where

``` js
css(
    StyleSheet.create({b: { color: null}}).b,
    StyleSheet.create({b: { color: 'blue'}}).b
)
```

results in `{color: {}}`

The reason was that `typeof null === 'object'`
Nothing like some good ol' Wat to make our lives interesting

Test Plan:
I added a test for this case in both recursive merge and css(x,y)
`npm test` shows all tests passing

### DIFF
--- a/src/util.js
+++ b/src/util.js
@@ -26,7 +26,8 @@ export const recursiveMerge = (a, b) => {
     // TODO(jlfwong): Handle malformed input where a and b are not the same
     // type.
 
-    if (typeof a !== 'object') {
+    // JSWat strikes again: typeof null === 'object'
+    if (a === null || typeof a !== 'object') {
         return b;
     }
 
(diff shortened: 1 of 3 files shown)


kevinbarabash:
@samiskin the output will include the 'null' if it appears second which is probably not what we want.

I think the following may be closer to what we want:
```
    if (a === null || typeof a !== 'object') {
      return b === null ? a : b;
    }
```

The reason why I didn't go with `return b || a;` is so that we can preserve order for numbers.

The tests should test both orders and there should also be a test for the ordering of numbers where one of the numbers is 0.

kevinbarabash:
The code for this has changed a lot and this no longer applies.

PR #107, opened 2016-07-22
add learn badge

kentcdodds:
So people can get a quick intro to how Aphrodite works and how to use it.

We don't have to use the badge. I'd be fine to put some text somewhere instead.

Also, I'm open to suggestions on topics for more videos. Big thanks to PayPal for letting me make these.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # Aphrodite: Inline Styles that work
 
-[![npm version](https://badge.fury.io/js/aphrodite.svg)](https://badge.fury.io/js/aphrodite) [![Build Status](https://travis-ci.org/Khan/aphrodite.svg?branch=master)](https://travis-ci.org/Khan/aphrodite) [![Gitter chat](https://img.shields.io/gitter/room/Khan/aphrodite.svg)](https://gitter.im/Khan/aphrodite)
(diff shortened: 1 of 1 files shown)


kentcdodds:
Here's what it looks like:

screen shot 2016-07-22 at 2 43 29 pm

jlfwong:
Oh cool - thanks for making these!

Who's the target audience for these? I personally find material like this much easier to absorb through documentation + (preferably editable) examples. I only briefly skimmed the material, but it seems like most of that is covered in the README?

In the past, I've generally found videos really helpful for things that are conceptually deep or have a lot of tradeoffs to explore (e.g. what makes this library better than that library), or for situations where there are a lot of sequential steps and it's not clear what your screen should look like at each interim step (e.g. 3D model manipulation in Blender).

For Aphrodite, I think that code samples (especially if accessible via a codepen/jsbin/whatever) are mostly self-explanatory if you're familiar with the underlying CSS concepts.

kentcdodds:
> Who's the target audience for these?

It's mostly introductory material. Stuff that you'd get from the README. But I think that there are many people who (like me) prefer to learn with video if possible rather than by reading docs. These are made like I make my Egghead.io videos: Short and focused.

> I've generally found videos really helpful for things that are conceptually deep or have a lot of tradeoffs to explore

These kinds of videos would definitely be really helpful. I'm not sure whether I'll personally make any like this. What I'd rather do is have a [tech chat](http://kcd.im/what-is-a-tech-chat) to cover that kind of thing and keep the videos focused and practical (could definitely do something like the sequential steps thing you're describing in a 6 minute video for example).

>  I think that code samples (especially if accessible via a codepen/jsbin/whatever) are mostly self-explanatory if you're familiar with the underlying CSS concepts.

Everyone learns differently, so I think that it's valuable to provide multiple methods for people to learn about how to use aphrodite. Also, each of the videos have code samples in the description, so that can be helpful as well.

jlfwong:
Okay, I'm happy enough to have this linked in the README.md, but would rather have it as a text link saying something along the lines of "If you'd rather watch introductory videos, you can find them here".

The main pain of videos like these is *updating* them, so if we have an API change, I'm inclined to remove it from the README unless someone has time to remake them. That said, Aphrodite's API is pretty small, so it seems unlikely.

kentcdodds:
Done: #109

PR #113, opened 2016-07-25
docs(tagline): update tagline

### LINKED ISSUE #111: Update tagline
See [this](https://twitter.com/jbscript/status/757567832536313858):

screen shot 2016-07-25 at 7 28 14 am

We need it to say something else. How about:

> CSS in JS, but with real CSS

kentcdodds:
closes #111 

We'd want to update the GitHub description as well.


kentcdodds:
ping

kentcdodds:
Ping on this again. It really is causing significant confusion in the community. [for example](https://medium.com/@oleg008/aphrodite-vs-jss-a15761b91ee3#.ch3fhwexk)

jlfwong:
That example doesn't seem like the author is confused about what Aphrodite does. "Aphrodite tries to make inline styles work by generating CSS." Seems accurate enough to me. The rest of the post seems pretty fair too, except that I disagree about a few of his argument, and disagree that Aphrodite is purely functional (`css()`, by design, has side effects).

I suppose the confusion comes from two different interpretations of "inline styles".

One is that the styles are conceptually inline in your JavaScript source code.
The other is that the styles end up inline the generated DOM nodes, as `<div style=...>`.

When you set the `style` property directly in React (or in vanilla JS), you're doing both. Aphrodite does only the first.

The idea of Aphrodite was to exactly mimic the behaviour of setting the `style` property, except with better performance and no restrictions on pseudo-selectors, media queries, etc.

So IMO, Aphrodite is indeed "inline styles", I suppose that term just means different things to different people.

I don't think your proposed tagline describes the behaviour any differently. Aphrodite isn't really "real css" any more than the styles presented in `<div style=...>` is.

kentcdodds:
Yeah, it was mostly this line:

> Aphrodite tries to make inline styles work by generating CSS.

Either way, I think most people think "inline styles" is referring to `<div style=...>` not `<div className=...>`. So when they hear aphrodite is "inline styles" they immediately think of all the shortcomings of inline styles:
- Hacks to get pseudo elements/states and media queries to work
- Perf issues with lots of elements
- Lack of vendor prefixing

When in reality, aphrodite has none of these things.

What if instead it were:

> CSS in JS, but by generating real CSS

or

> CSS in JS, without the shortcomings of inline styles

That second one is my favorite.

README.md
```diff
@@ -1,4 +1,4 @@
-# Aphrodite: Inline Styles that work
+# Aphrodite: CSS in JS, but with real CSS
```
lencioni: The description on GitHub has since been updated to "Framework-agnostic CSS-in-JS with support for server-side rendering, browser prefixing, and minimum CSS generation" (#259). @kentcdodds if you update your PR to match I'd be happy to merge it in!

kentcdodds: Done!

PR #114, opened 2016-07-25
Add a caveat about how to combine styles

xymostech:
This comes up a lot in different discussions, and there isn't a definitive
reference for it, so I figured I should add one! No more lengthy explanations
in github issues.

I'm very open to feedback about the wording of this, please leave comments!

@jlfwong @kentcdodds @JedWatson

### DIFF
--- a/README.md
+++ b/README.md
@@ -264,6 +264,92 @@ The generated css will be:
   }
 ```
 
+## Overriding styles
+
+When combining multiple aphrodite styles, you are strongly recommended to merge all of your styles into a single call to `css()`, and should not combine the generated class names that aphrodite outputs (via string concatenation, `classnames`, etc.).
+For example, if you have a base style of `foo` which you are trying to override with `bar`:
+
+### Do this:
+
+```js
+const styles = StyleSheet.create({
+  foo: {
+    color: 'red'
+  },
+
+  bar: {
+    color: 'blue'
+  }
+});
+
+// ...
+
+const className = css(styles.foo, styles.bar);
+```
+
+### Don't do this:
+
+```js
+const styles = StyleSheet.create({
+  foo: {
+    color: 'red'
+  },
+
+  bar: {
+    color: 'blue'
+  }
+});
+
+// ...
+
+const className = css(styles.foo) + " " + css(styles.bar);
+```
+
+Why does it matter? Although the second one will produce a valid class name, it cannot guarantee that the `bar` styles will override the `foo` ones.
+The way the CSS works, it is not the *class name that comes last on a element* that matters, it is specificity. When we look at the generated CSS though, we find that all of the class names have the same specificity, since they are all a single class name:
+
+```css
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -248,6 +248,92 @@ The generated css will be:
   }
 ```
 
+## Overriding styles
+
+When combining multiple aphrodite styles, you **must** merge all of your styles into a single call to `css()`, and **must not** combine the generated class names that aphrodite outputs (via string concatenation, `classnames`, etc.).
```
kentcdodds: Using the word `must` seems inflexible. I'd use "`should`" or "`it is recommended to`"

kentcdodds:
LGTM (other than that one note). Thanks for adding this!

README.md
jlfwong: "must" has a specific meaning in RFC style, but I agree that it might read as a little forceful here.

JedWatson:
Looks good to me as well.

Thinking about that, `classnames` has been shockingly popular, I can imagine anyone switching to Aphrodite missing it if they're used to that way of applying conditional classes. Would you be interested in investigating a framework-specific port of it?

e.g

```jsx
import { css, classNames } from 'aphrodite'

<div className={classNames(styles.foo, { styles.bar: cond })} />
```

Would make this whole explanation friendlier ("don't use the classnames package, import it from Aphrodite" rather than "you can't do this")

kentcdodds:
I like what @JedWatson suggests because I've often found myself having to do fancy stuff when calling `css` when the styles are conditional. This would simplify things quite a bit.

jlfwong:
You shouldn't need to do fancy stuff, since if any of the arguments is false-y, it gets ignored.

Instead of `<div className={classNames(styles.foo, { styles.bar: cond })} />`, you can do this:

`<div className={classNames(styles.foo, cond && styles.bar })} />`

This is probably not documented in the README.md though, if someone would like to add that :)

kentcdodds:
Would this work?

```jsx
<div className={classNames(styles.foo, cond && styles.bar, cond2 && styles.baz, cond3 && styles.foobar })} />
```

jlfwong:
@kentcdodds whoops, sorry -- I miswrote the comment about "you shouldn't need to do fancy stuff" -- see edit. My point was that this kind of conditional application of classnames is already supported by the `css()` function. I was not discussing the semantics of a potential `classnames` function, I was discussing the existing semantics of the `css()` function.

kentcdodds:
Haha, neither was I actually! See _my_ update :-)

xymostech:
@kentcdodds yes, that does work! +1 to adding documentation about this, I didn't realize people didn't know this.

jlfwong:
I think this just slipped through the cracks? LGTM to merge @xymostech

xymostech:
(I made the wording a little less intense. Merge incoming...)

PR #116, opened 2016-07-27
add conditional css API example

kentcdodds:
ref: https://github.com/Khan/aphrodite/pull/114#issuecomment-235669930

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,6 +79,22 @@ const styles = StyleSheet.create({
 });
 ```
 
+## Conditionally Applying Styles
+
+Note: If you want to conditionally use styles, that is simply accomplished via:
+
+```jsx
+const className = css(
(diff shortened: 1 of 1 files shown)


xymostech:
Thank you for adding this! Some comments:

- I would add a note like "any falsey arguments will be ignored. So, styles can be applied conditionally..." to make it clear why this works.
- Could you put each condition on a separate line so the example is easier to read and so it isn't too wide?
- You could just make this into a section, like "Conditionally applying styles" or something

kentcdodds:
Great feedback. Updated 👍

kentcdodds:
Here's what it looks like:

screen shot 2016-07-27 at 1 29 48 pm

xymostech:
LGTM! @jlfwong any thoughts?

jlfwong:
Thanks!

Issue #100, opened 2016-07-08
prefixed versions of rules are appearing after non-prefixed rules in generated styles


danhayden:
Sample output:

``` css
.wrapper_5h5b2v {
    flex: 1 1 -moz-calc(100% - 300px);
    flex: 1 1 -webkit-calc(100% - 300px);
    flex: 1 1 calc(100% - 300px);
    -webkit-flex: 1 1 calc(100% - 300px);
    -ms-flex: 1 1 calc(100% - 300px);
}
```

As you can see in the the image below from Chrome dev tools this is causing non-prefixed rules to be overwritten in the browser by prefixed ones

![screen shot 2016-07-08 at 12 57 08](https://cloud.githubusercontent.com/assets/1385253/16686669/9deebd64-450b-11e6-9a95-db29f4fdfebf.png)

kentcdodds:
Genuinely curious here. What's the problem with having the vendor-prefix version overwrite the non-prefix version? THe result is the same right? I'm probably missing something.

danhayden:
The vendor-prefix versions are not always the same as the final W3C compliant implementation.
Sometimes they can be buggy due to experimental, browser-specific code and may deviate from the specification.

xymostech:
@danhayden Thanks for the report! I wasn't really convinced by this, but I found [this CSS Tricks article](https://css-tricks.com/ordering-css3-properties/) explaining how `border-radius` and `-webkit-border-radius` act differently, so I definitely believe this.

It sounds like we need to ensure that the unprefixed versions come last. I think this might be most easily accomplished with a change to the auto-prefixer?

elrumordelaluz:
Seems that [inline-style-prefixer](https://github.com/rofrischmann/inline-style-prefixer) generate the `-prefixed` styles in the right order, also when values are `Arrays` like in the following code, where the _un-prefixed_ is the last. 

```js
{
  // ...
  display: ['-webkit-box', '-moz-box', '-ms-flexbox', '-webkit-flex', 'flex'],
  //...
}
```

> It's preferable to have the W3C implementation used when it's available, to help ensure consistency between all the browsers that support it.

---

### Autoprefixer

Maybe this is for a separate issue, but take the opportunity to say that switch to [autoprexier](https://github.com/postcss/autoprefixer) should be pretty useful, since allows to control the _userAgent_ by queries provided by [browserslist](https://github.com/ai/browserslist).

[...]

PR #121, opened 2016-08-14
Move vendor prefixed rules before unprefixed ones

elrumordelaluz:
This change attempt to solve #100.


elrumordelaluz:
Closed doe to make a cleaner commit. I'm sorry.

PR #122, opened 2016-08-14
Move vendor-prefixed rules before unprefixed

elrumordelaluz:


jlfwong:
Hi @elrumordelaluz,

Thanks for taking this on. I think this implementation ends up doing more work, and is a little bit less efficient than #123. Would you mind taking a look at that implementation and see if it accomplishes the same goals as this?

elrumordelaluz:
@jlfwong you are absolutely right. Work as expected.

Didn't come in mind to sort the `Array` before `flatten` it. In fact it's pretty easy `sort()`ing since all _prefixes_ begin with a `-`.

> It's not clear to me why exactly we were doing this, but inline-style-prefixer seems to do the right thing.

In case you are asking why we need _prefixed_ rules before, it's a best practice to *always place the standard versions last* to let the _cascade_ take care which wins, in case the browser supports the _unprefixed_, solved. 

I copied a excerpt where this (and other) stuff is explaned, from the awesome book of @leaverou [here](https://gist.github.com/elrumordelaluz/77b75762526d986b1b51c07a367670af)

PR #123, opened 2016-08-15
Sort properties to ensure prefixed properties precede unprefixed ones

### LINKED ISSUE #100: prefixed versions of rules are appearing after non-prefixed rules in generated styles

jlfwong:
I also changed the code to no longer sort the values when inline-style-prefixer
generates multiple values. It's not clear to me why exactly we were doing this,
but inline-style-prefixer seems to do the right thing. See #71 for original
context.

Fixes #100

### DIFF
--- a/src/generate.js
+++ b/src/generate.js
@@ -140,39 +140,37 @@ export const generateCSSRuleset = (selector, declarations, stringHandlers,
     const handledDeclarations = runStringHandlers(
         declarations, stringHandlers);
 
-    const prefixedDeclarations = prefixAll(handledDeclarations);
+    const prefixedDeclarations = objectToPairs(
+      prefixAll(handledDeclarations)
+    ).map(([key, val]) => [key, kebabifyStyleName(key), val]);
 
-    const prefixedRules = flatten(
-        objectToPairs(prefixedDeclarations).map(([key, value]) => {
+    // After we've applied prefixing, we now have an array of [key,
+    // kebabKey, value] triples, where the kebabKey is the maybe
+    // vendor-prefixed css property name (e.g."transition" or
+    // "-webkit-transition")
+    //
+    // We want to sort this list of triples to ensure that vendor prefixed keys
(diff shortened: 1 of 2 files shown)


jlfwong:
I don't *think* this is a breaking change from a normal user perspective, though it does change the output (which I don't consider part of the API, so long as it accomplishes the same thing).

zgotsch:
LGTM. This should probably bump whatever version number Aphrodite associates with an API change, since it could cause breakage, even if it's a browser's fault.

zgotsch:
>  It's not clear to me why exactly we were doing this,
but inline-style-prefixer seems to do the right thing

We were seeing different orderings of properties between travis and my local -_-. I think inline-style-prefixer may have fixed some things since the original implementation on our side.

xymostech:
This looks fine to me! It doesn't seem like this needs a breaking change to me, but better safe than sorry I guess.

Speaking of breaking changes, do you want to bump to v1.0.0 some time? Semver says

> How do I know when to release 1.0.0?

> If your software is being used in production, it should probably already be 1.0.0. If you have a stable API on which users have come to depend, you should be 1.0.0. If you’re worrying a lot about backwards compatibility, you should probably already be 1.0.0.

danhayden:
Any ETA for this being merged?

limitedmage:
Is this ever getting merged?

xymostech:
This was fixed by v2.0.5 of inline-style-prefixer, so I'm going to close this.

PR #125, opened 2016-08-16
add EditorConfig for happiness :)

jokeyrhyme:
- based on a catch-all EditorConfig file I have here: https://github.com/jokeyrhyme/standard-editorconfig

- uses Google's style guide for HTML, so updated the 1 HTML file to match

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,43 @@
+# EditorConfig is awesome: http://EditorConfig.org
+
+# https://github.com/jokeyrhyme/standard-editorconfig
+
+# top-most EditorConfig file
+root = true
+
+# Unix-style newlines with a newline ending every file
+[*]
+end_of_line = lf
+insert_final_newline = true
+
+# Set default charset
+[*]
+charset = utf-8
+
+# Other good defaults
+[*]
+indent_size = 2
+indent_style = space
+trim_trailing_whitespace = true
+
+# HTML
+[*.{htm,html}]
+indent_size = 4
+indent_style = space
+trim_trailing_whitespace = true
+
+# JavaScript, JSX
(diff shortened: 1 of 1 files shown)


kevinbarabash:
@jokeyrhyme we probably want to use Khan Academy's style guide which uses 4 spaces indents for JavaScript, see https://github.com/Khan/style-guides/blob/master/style/javascript.md#indentation.  We don't spell out indent sizes for HTML or YAML, but I think 4 and 2 respectively is probably the way to go.

jokeyrhyme:
@kevinbarabash thanks for the feedback! I've changed the white-space rules as follows:
- JS: 4 spaces
- HTML: 4 spaces
- YAML: 2 spaces

Let me know if you have other concerns. Cheers!

kevinbarabash:
@jokeyrhyme LGTM.

jokeyrhyme:
I have submitted the CLA form as well, as part of my other PR.

jokeyrhyme:
Oh, @kevinbarabash I just wanted to point out that this EditorConfig sets indenting to 4 spaces for JS, JSX and JSON files. We'd definitely discussed JS files, but I just wanted to double check the others.

kevinbarabash:
@jokeyrhyme thanks for checking.  I don't think there are any files with the .jsx extension.  There are some .js files with jsx markup in them and those are 4 space indents.  The .json files on the other hande seem to be 2 space indents so we should probably maintain that.

jokeyrhyme:
@kevinbarabash I've changed the EditorConfig so that JSON files have the 2-space indenting. JS and JSX files are unchanged (4-space).

xymostech:
This LGTM! Thanks @jokeyrhyme!

PR #126, opened 2016-08-16
WIP: add ESLint with Semi-Standard and React rules; fix compliance

jokeyrhyme:
- I noticed that most of the code was 4-space-indenting, but did also see a few cases of 2-space-indenting, so I went with 2-space which is what "Standard" recommends

    - https://github.com/feross/standard

    - https://github.com/feross/eslint-config-standard

- I noticed that you use semicolons, so I pulled in the Semi-Standard rules as well

    - https://github.com/Flet/eslint-config-semistandard

    - embed these rules, because the npm package has peerDependency issues for now

- most of the changes in the second commit are generated automatically by `eslint --fix .`

- example directory also uses

    - https://github.com/feross/standard-react

    - https://github.com/feross/eslint-config-standard-react


package.json
```diff
@@ -42,6 +44,12 @@
     "babel-core": "^5.8.25",
     "babel-loader": "^5.3.2",
     "chai": "^3.3.0",
+    "eslint": "^3.3.1",
+    "eslint-config-standard": "^5.3.5",
+    "eslint-config-standard-react": "^3.0.0",
+    "eslint-plugin-promise": "^2.0.1",
+    "eslint-plugin-react": "^6.1.1",
+    "eslint-plugin-standard": "^2.0.0",
```
kevinbarabash: I don't understand the need for the `-react` and `-promise` plugins.

kevinbarabash:
@jokeyrhyme thanks for tackling one of the TODOs in the README.md.  Can you update the eslint configure to follow existing conventions in order to minimize changes.  In particular it looks like we're using:
- 4 space indents
- double quotes
- no space between `function` and `(`

package.json
jokeyrhyme: It's required by Standard: https://github.com/feross/eslint-config-standard/blob/master/package.json#L15

jokeyrhyme:
With just [Standard](https://github.com/feross/standard) and [Semi-Standard](https://github.com/Flet/semistandard) rules, I agree that there are way too many changes:

- `16 files changed, 1280 insertions(+), 1285 deletions(-)16 files changed, 1280 insertions(+), 1285 deletions(-)`

Standard + Semi-Standard + 4-space indents:

- `16 files changed, 308 insertions(+), 312 deletions(-)`

Standard + Semi-Standard + 4-space indents + no space before `function` paren:

- `15 files changed, 291 insertions(+), 294 deletions(-)`

Standard + Semi-Standard + 4-space indents + no space before `function` paren + double-quotes:

- `17 files changed, 401 insertions(+), 403 deletions(-)`

- Unless you really want it, I'm going to leave out the double-quotes rule, as it actually results in more changes. It would seem the code-base is primarily single-quotes for now.

This is still a work-in-progress while I narrow down the other rules required to minimise changes.

jokeyrhyme:
I'm having an issue with multi-line array literals and [ESLint's indent rule](http://eslint.org/docs/rules/indent)

For some reason, even with 4-space indents configured, it is expecting array literals to be indented with 2 spaces. I've tried explicitly enabling the "MemberExpression" option for the indent rule, which is the closest thing I could think of, but that doesn't have any effect in this case.

There's an array literal here that is tripping this up: https://github.com/Khan/aphrodite/blob/master/examples/src/StyleTester.js#L24

Any advice? It seems weird to have 4-space indents everywhere except for multi-line array literals.

jokeyrhyme:
@kevinbarabash see #153 for what is hopefully an easier to review PR

kevinbarabash:
@jokeyrhyme thanks.  I'm going close this PR and review the other one.

PR #132, opened 2016-08-29
Added promise support to renderStatic

gtg092x:
I'm yielding on the rendering of my app until all of my data requirements have been loaded. I want to keep the buffer open while this is happening, though.

This means that I need to wait on an async resolution before I can do my final static render, these changes should make it so my scenario is supported.

I have not bumped the npm version, if that's needed I'll be happy to make that change.

Thanks for all your great work on aphrodite!


xymostech:
If you're just going to do one big `ReactDOMServer.renderToString()` at the end, is there a reason you need for Aphrodite to change? (just trying to understand why this is needed, I would rather not add something if it can be accomplished with the current API)

Copying from your example, could you not just do...

```
return new Promise(resolve => {
    resolve(StyleSheetServer.renderStatic(
        () => ReactDOMServer.renderToString(<App/>)));
}).then(({html, css}) => {
    ...
});
```

gtg092x:
I've got to load my app and resolve some data resources before I can render it to string. My code would like like this:

```js
StyleSheetServer.renderStatic(
  // this uses a promise to load my app and crawl data requirements
  () => ResolveDependenciesForApp(<App />) 
).then(({html, css}) => {
    // app has injected data
});
```

If I try to load my app outside of the renderStatic promise, I won't have a buffer set and will get the document error. I need to load the app to resolve data requirements, though.

xymostech:
I'm confused why you are getting an error about a buffer being set if you haven't yet tried to render things. Are you calling `css()` in places other than inside your render functions?

gtg092x:
I need to do an initial load of my app without hydrated data. Rendering to
string is synchronous, but I won't be doing it immediately because data
isn't available yet.

The only scenario you currently support is having all data loaded before I
enter the app, which isn't my case. I'm doing an initial load, gathering
resources, and rendering to string when it's ready.

On Aug 29, 2016 3:00 PM, "Emily Eisenberg" <notifications@github.com> wrote:

I'm confused why you are getting an error about a buffer being set if you
haven't yet tried to render things. Are you calling css() in places other
than inside your render functions?

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/Khan/aphrodite/pull/132#issuecomment-243271043>, or mute
the thread

gtg092x:
@xymostech would it help if I gave a clearer use case? I'm trying a server side rendering approach where I don't render to string on the initial react render, but wait until the app is ready. I don't think that's possible now with aphrodite.

lvauvillier:
Same problem using apollostack.

renderToStringWithData() return a promise
http://dev.apollodata.com/react/server-side-rendering.html#renderToStringWithData

jlfwong:
Ah, interesting. That use case is unfortunate, and I don't have a work around, but having `StyleSheetServer.renderStatic` return a promise isn't going to work without a dramatic retooling of Aphrodite's API.

In your tests, it appears to work just fine, but if you were running this on a server with multiple concurrent connections, it would fail.

From your example:

```
StyleSheetServer.renderStatic(
  // this uses a promise to load my app and crawl data requirements
  () => ResolveDependenciesForApp(<App />) 
).then(({html, css}) => {
    // app has injected data
});
```

Consider what happens if this is asynchronous and you have 2 concurrent requests.

Request 1: `StyleSheetServer.renderStatic` called, `ResolveDependenciesForApp` called, performing HTTP request 1 and rendering `<App/>`, resulting in many `css()` calls. Response pending.
Request 2: `StyleSheetServer.renderStatic` called, `ResolveDependenciesForApp` called, performing HTTP request 2 and rendering `<App/>`, resulting in many `css()` calls. Response pending.
Request 1: HTTP request 1 responds. `renderStatic` promise resolves. The CSS buffer is flushed, containing the CSS from *both request 1 and request 2*.
Request 2: HTTP request 2 responds. `renderStatic` promise resolves. The CSS buffer is flushed, containing *empty CSS*.

But let's try to solve your original problem. What is `ResolveDependenciesForApp` doing? Are you doing a full `render()` of your app, then crawling through the virtual DOM tree?

I'm closing this because having `renderStatic` be asynchronous fundamentally won't work since `css()` relies on global state so it can't deal with 2 concurrent requests. But I am interested in understanding your problem.

Other solutions work around this by storing this state in `context` that's shared between all components in a subtree, but that would request `css` to have access to the React component, and Aphrodite doesn't depend on React at all.

gtg092x:
Thank you for taking the time to provide feedback here. I'll see if there's a way to follow this up with similar functionality while scoping each render to request.

vlasenko:
Hi there,

Thank you for the awesome stylesheet library. 

@jlfwong I would like to raise the point that supporting callback returning `Promise` might be legitimate case when using `StyleSheetServer.renderStatic`. For example one has this case if it uses server-side rendering with ApolloStack:
https://github.com/apollostack/react-docs/blob/master/source/server-side-rendering.md#user-content-getDataFromTree

Right now to support both ApolloStack server-side rendering and Aphrodite styles one have to use hacks like this:
https://github.com/sysgears/apollo-fullstack-starter-kit/commit/b7db7c6b550026149cec4734d12291456b4b190e

Thanks,
Victor

xymostech:
Okay, I finally got around to digging in to what was happening here. It looks like @jlfwong was right, the reason you're getting errors here is that you're doing a full render of your tree during your `ResolveDependenciesForApp` step (which happens in here: https://github.com/apollostack/react-apollo/blob/af59d099972799bb845892b313df35793909cd38/src/server.ts#L35).

It actually looks like maybe the _render_ happens synchronously (or at least the first render, it like there's some recursion in there that might cause async renders) despite the data fetching being async. Just setting up your apollo starter kit, I was able to get things to work by doing 
```js
StyleSheetServer.renderStatic(() => getDataFromTree(component)).html.then(context => {
```
but that's an awful hack that probably won't work for everyone.

In general though, I think that adding a function that works for async renders is probably smart. Would having a function like `StyleSheetServer.renderStaticAsync` that looks like
```js
      let context;
      StyleSheetServer.renderStaticAsync(() => {
        return getDataFromTree(component).then(context_ => {
          context = context_;
          return ReactDOM.renderToString(component);
        });
      }).then(({ html, css }) => {
```
sound reasonable?

jlfwong:
I'm still a bit confused as to how we can possibly write safely write server rendering code that might have multiple concurrent requests if we have an API that works like this.

Maybe I'm missing something major, but as far as I can tell, our API of the `css()` function is fundamentally incompatible with supporting multiple concurrent async requests (i.e. if you have two interleaved calls to `StyleSheetServer.renderStaticAsync`, you're going to get bad results in one of them). See above comment for the more detailed explanation of why I think the paradigm won't work.

vlasenko:
@xymostech This hack was used before, but it doesn't work all the time, since promises are used inside the recursion, that you've seen. And to solve this, I had to start and stop buffering manually:
https://github.com/sysgears/apollo-fullstack-starter-kit/issues/8

`StyleSheetServer.renderStaticAsync` looks good, thanks!

vlasenko:
@jlfwong I think its possible to detect call of `StyleSheetServer.renderStaticAsync` inside another `renderStaticAsync` or inside `renderStatic` and warn user on the console, that he might be using them wrong way.

jlfwong:
@vlasenko I agree it's possible to write a server in a way in which this would be safe, but that safe way is not the way the vast majority of Node.js servers work. I'm wary of providing an API that would be used wrong 95% of the time.

I guess we could make it throw an exception if the calls interleave, but I suspect that all of the use cases that people are interested in would throw then.

This pattern would only work if you only allow one request to be processed at a time, which is not, AFAIK, the normal processing method of any Node.js server that does async communication. Single threaded concurrent request handling is one of the major value props of Node.js over other HTTP server systems. Other servers get around problems like this by not doing evented IO, meaning the handling of multiple requests do not interleave.

If there's some subtletly I'm missing here, please let me know!

xymostech:
@jlfwong Ah yes, you are correct. I forgot that other renders might be happening concurrently. Hmm. I wonder if other styling libraries have figured this out. Cursory glances at [glamor's API](https://github.com/threepointone/glamor/blob/master/docs/server.md), and [jss's API](https://github.com/cssinjs/jss/blob/master/docs/ssr.md) says they have the same problem. :(

It might be possible to create a React-specific solution, where we associate a set of buffered styles with a given React tree, and then buffer into that set when we're rendering something from that tree? That sounds pretty scary though.

@vlasenko For your specific use case, it looks like you do a render of the tree synchronously and then async call the loading functions. If you separate those out, you could wrap the render in a `StyleSheetServer.renderStatic()` and then do the tree traversal/async loading separately?

vlasenko:
@xymostech I'm not sure I fully understand you. Could you show me rough code of proposed solution in my case?

xymostech:
@vlasenko I guess this more has to do with the internals of how react-apollo works than the stuff we were talking about. What I meant was that in here: https://github.com/apollostack/react-apollo/blob/8cb978bdc8cd7b59ed2cf6accd7ca01ff627d7d9/src/server.ts#L92 there are two steps that happen:

 1. The react tree gets rendered (in the `walkTree` function), and accumulates data fetching functions
 2. Those data fetching functions are then run and accumulated into a final result (during `Promise.all(mappedQueries)` in `getDataFromTree`)

Adding promise support via `StyleSheetServer.renderStaticAsync` would allow you to accumulate aphrodite styles from steps 1. and 2., and flush the styles once both of those things have happened, but would also introduce problems with concurrency like @jlfwong said.

A better solution would be to separate out parts 1. and 2., wrap 1. in a call to `StyleSheetServer.renderStatic`, and _then_ perform 2. separately.

Does that make more sense?

vlasenko:
@xymostech Yes, it makes more sense, thanks. But I guess, it cannot be solved this way for `react-apollo`. 

I think the things are working this way:
`react-apollo` walks the React tree and gathers GraphQL queries, then it runs these queries. These queries in turn might change resulting React tree, for example imagine list rendering, when after running query new list elements appear. Thats why `react-apollo` need to recursively call `getDataFromTree` to handle these new elements that appear AFTER running queries:
https://github.com/apollostack/react-apollo/blob/8cb978bdc8cd7b59ed2cf6accd7ca01ff627d7d9/src/server.ts#L128

That's why splitting Tree walking and running queries is not really possible.

However there is one nuisance. During running `getDataFromTree` we are not interested in CSS styles generation at all. We just need Aphrodite to not throw exceptions during this phase. Since true rendering step, which will be sync will happen after `getDataFromTree`.

See the code here:
https://github.com/sysgears/apollo-fullstack-starter-kit/blob/b7db7c6b550026149cec4734d12291456b4b190e/src/server/middleware/website.jsx#L47-L57

The true rendering happens on line 57, at this point the React tree is stabilized and we can do sync rendering.

At lines 49, 50 and 53 I'm just trying to tell `Aprodite`, hey don't worry, just ignore styles for now. Perhaps `Aphrodite` should have not `StyleSheetServer.renderStatic`, but `StyleSheetServer.ignoreStyles` which takes Promise, instead? I'm not sure how best to solve this. But I'm sure our concern is more specific, it is not handle async rendering, but rather ignore generated styles.

Thoughts?

Victor

jlfwong:
Ignoring generated styles has exactly the same problem. If you start ignoring styles in another request, and want those styles to be collected in a concurrently running request, it won't work.

jlfwong:
FWIW, supporting async rendering like this is certainly *possible*, just not with Aphrodite's current API.

If someone is interested in forking Aphrodite, making it depend on React, then changing the api to use `this.css()` instead of `css()`, and stashing the state in the `context` of the React tree, like @xymostech mentioned:

> It might be possible to create a React-specific solution, where we associate a set of buffered styles with a given React tree, and then buffer into that set when we're rendering something from that tree? That sounds pretty scary though.

But I don't think that change belongs in this repository.

I'm going to re-close this issue until someone can provide a plan that (1) doesn't run into the problems with concurrency described above, and (2) doesn't dramatically change Aphrodite's API by e.g. making it depend on React. I'm fairly confident that there is no plan that satisfies both (1) and (2), but I've been wrong about plenty of things I've been confident about!

vlasenko:
After experimenting and thinking about this problem more, I believe the workaround exists which will work in most practical cases, with current Aphrodite's API. 

Styles must not be collected during async data loading phase, they should be collected only during rendering phase. This approach doesn't have problems in practice, because during server side rendering of HTML response, another server side rendering of HTML response cannot occur, because there is no blocking IO during HTML rendering. The phrase blocking IO here is used in the sense of blocking operations:
https://nodejs.org/en/docs/guides/blocking-vs-non-blocking/

Here is the solution I came up with React Apollo and Aphrodite:
https://github.com/sysgears/apollo-fullstack-starter-kit/blob/1ec149adb24ebf0f5c66847d3257386c706d588d/src/server/middleware/website.jsx#L49-L56

zaklampert:
@vlasenko how do you get around the fact that `StyleSheetTestUtils` is null in NODE_ENV production?

vlasenko:
@zaklampert This question should not really be addressed to me. At the time my comment was written it was May 2017, and `StyleSheetTestUtils` was available in current version of `aphrodite` at that time for NODE_ENV = production. Yeah, it was not a nice way to work around promises support in `renderStatic`, but it worked. Now it is up to `aphrodite` developers to understand that when you do server side rendering and use promises they can perfectly support this use case if there is no blocking IO during this phase. And they can bake in support for this use case without much trouble.

PR #112, opened 2016-07-25
Use JSX syntax highlighting for JSX code in readme

lencioni:
This improves the syntax highlighting of this code block.

PR #138, opened 2016-09-10
Improve syntax highlighting in readme code snippet

colinwren:
## Before (Using `jsx` syntax highlighting):
screen shot 2016-09-10 at 12 49 37
## After (Using `js` syntax highlighting):
screen shot 2016-09-10 at 12 49 25


xymostech:
:P You and the person who switched it to jsx in #112 will have to fight it out I suppose?

More seriously, neither of them look terribly great, but the code is actually JSX, so I'm inclined to stick with the JSX styling and hope that it gets better highlighting support in the future. Thanks for the PR though!

PR #142, opened 2016-09-17
Add a RunKit example to Aphrodite

### LINKED ISSUE #140: Add a RunKit example
I noticed that there's a way to add a default example to the RunKit (formerly TonicDev?) example that for a repository: https://discuss.runkit.com/t/customizing-your-npm-packages-page-on-tonic/23

It would be cool if we added one of these for Aphrodite! I came up with a server-side rendering example that we could use: https://runkit.com/xymostech/aphrodite-server-side-rendering-example I tried to make a non-server-side-rendering example but it required jsdom and a `setTimeout` which was pretty ugly. 

Thoughts?

xymostech:
Summary: RunKit (previously TonicDev) lets you add a custom example on
the package page linked to from npmjs.com:
https://discuss.runkit.com/t/customizing-your-npm-packages-page-on-tonic/23

This adds such an example!

Closes #140

Test Plan:
 - Publish this as the new package, `aphrodite-test`
 - See that the runkit integration works :)

### DIFF
--- a/examples/runkit.js
+++ b/examples/runkit.js
@@ -1,0 +1,33 @@
+// Aphrodite server-side rendering example
+
+// Make a component that generates some styles and returns some HTML.
+function render() {
+    const {StyleSheet, css} = require("aphrodite/no-important");
+
+    // Make some styles
+    const styles = StyleSheet.create({
+        red: {
+            color: "red",
+
+            ":hover": {
+                color: "blue",
+            },
+        },
+    });
+
+    // Generate some CSS with Aphrodite class names in it.
+    return `<div class=${css(styles.red)}>
+        Hover, and I'll turn blue!
+    </div>`;
+}
+
+const {StyleSheetServer} = require("aphrodite");
+
+// Call our render function inside of StyleSheetServer.renderStatic
+const {css, html} = StyleSheetServer.renderStatic(() => {
+    return render();
+});
+
+// Observe our output HTML and the Aphrodite-generated CSS
+console.log("Output HTML:", html);
+console.log("Output CSS:", css.content);
--- a/package.json
+++ b/package.json
@@ -52,5 +52,6 @@
   "dependencies": {
     "asap": "^2.0.3",
     "inline-style-prefixer": "^2.0.0"
-  }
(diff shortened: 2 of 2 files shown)


xymostech:
https://runkit.com/npm/aphrodite-test

examples/runkit.js
```diff
@@ -0,0 +1,33 @@
+// Aphrodite server-side rendering example
+
+// Make a component that generates some styles and returns some HTML.
+function render() {
+    const {StyleSheet, css} = require("aphrodite/no-important");
```
jlfwong: Seems weird to use the non-default option in the example

xymostech: :P If a console statement is too long in RunKit, it adds an ellipse, which I thought was annoying, so I went with this to fix that. I'm happy to undo that though!

image

xymostech:
> Did you mean to check in dist/aphrodite.js? Looks like those changes have nothing to do with this review.

updated the pr, I had a git pre-commit hook which automatically built the files from back when we always committed built files. I removed it now :P

examples/runkit.js
jlfwong: Hehe, sad, but okay!

jlfwong:
LGTM!

xymostech:
Can you accept the review? I can push it anyway but I want to see what it looks like when you do :)

xymostech:
Also, I brought this up a bit ago but now that we have new things to land, I'll bring it up again: Do we want to move to version 1.0 next time we release? Semver says

> If your software is being used in production, it should probably already be 1.0.0.

and Aphrodite is very in production.

kentcdodds:
Yes please :)

jlfwong:
@xymostech Was going to suggest we make 1.0.0 the version with plugins, though perhaps it's weird to do the bump to 1.0.0 when introducing a somewhat experimental feature? I don't have strong feels, and would be happy for us to bump immediately if you think that's more appropriate.

xymostech:
Oh yeah, we can do it after that lands. Just making sure you're on board with doing it sometime soon :)

PR #146, opened 2016-09-22
:memo: we support animations

basarat:
Did a search on the page and thought perhaps it wasn't supported. Googled, found answer so sent this PR :rose:

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,6 +8,7 @@ Support for colocating your styles with your JavaScript component.
 - Supports media queries without window.matchMedia
 - Supports pseudo-selectors like `:hover`, `:active`, etc. without needing to
   store hover or active state in components. `:visited` works just fine too.
+- Supports keyframes using `animation` property
 - Supports automatic global `@font-face` detection and insertion.
 - Respects precedence order when specifying multiple styles
 - Requires no AST transform


kentcdodds:
Awesome :clap:

jlfwong:
Hi @basarat!

That is indeed a hole in our docs!

It would be great if you could add a section right below "Font Faces" about how this feature is used.

You can find an example usage here: https://github.com/Khan/aphrodite/blob/master/tests/inject_test.js#L352

Thanks!

kentcdodds:
Also relevant :) https://www.youtube.com/watch?v=BIpKF_CMhu8&list=PLo4Zh55ZzNSBP78pCD0dZJi9zf8CA72_M&index=4

xymostech:
@basarat Awesome! We use the `animationName` property actually. :)

basarat:
> We use the animationName property actually

I think is `animation: keyframes, animationName: nameStringUsedToRegisterTheKeyframes`

xymostech:
> I think is animation: keyframes, animationName: nameStringUsedToRegisterTheKeyframes

You can read the source yourself, we only use the `animationName` property: https://github.com/Khan/aphrodite/blob/master/src/inject.js#L60

basarat:
> we only use the animationName property:

Yup. My bad. 

Closing to let some other hero take it over :rose:

PR #154, opened 2016-10-07
RFC: Accept arrays of styles in css()

ide:
Accepting both styles and arrays of styles improves developer ergonomics with how I use Aphrodite when passing styles into components. The biggest problem I run into is that the spread operator doesn't work with undefined, so this code fails:

```js
// OwnerComponent.js
class OwnerComponent extends React.Component {
  render() {
    return <Component />;  // we're happy with the default styles
  }
}
```
```js
// Component.js
class Component extends React.Component {
  static propTypes = { styles: PropTypes.array };
  render() {
    // this css() call fails because spreading `undefined` is an error
    return <div className={css(styles.myStyle, ...this.props.styles)} />;
  }
}

const styles = StyleSheet.create({...});
```

The second, more cosmetic issue, is that even if I'm passing down just one style I need to create an array:

```js
// OwnerComponent.js
class OwnerComponent extends React.Component {
  render() {
    return <Component styles={[styles.customStyle]} />;  // need the [ ]
  }
}

const styles = StyleSheet.create({...});
```

With this PR, you'd write `css(styles.myStyle, this.props.styles)` and it'd address both these issues. Aphrodite would flatten `this.props.styles` if it were an array, it would leave it alone if it were an Aphrodite style object, and it would also leave it alone and filter it out if it were falsy.

As a data point, React Native flattens style arrays (recursively, actually) and it's pretty convenient with little downside. So just wanted to float this idea out there.

Test Plan: added unit tests

### DIFF
--- a/README.md
+++ b/README.md
@@ -95,6 +95,41 @@ const className = css(
 
 This is possible because any falsey arguments will be ignored.
 
+## Combining Styles
+
+To combine styles, pass multiple styles or arrays of styles into `css()`. This is common when combining styles from an owner component:
+
+```jsx
+class App extends Component {
+    render() {
+        return <Marker styles={[styles.large, styles.red]} />;
+    }
+}
+
+class Marker extends Component {
+    render() {
+        // css() accepts styles, arrays of styles (including nested arrays),
+        // and falsy values including undefined.
+        return <div className={css(styles.marker, this.props.styles)} />;
+    }
+}
+
+const styles = StyleSheet.create({
+    red: {
+        backgroundColor: 'red'
+    },
+
+    large: {
+        height: 20,
+        width: 20
+    },
+
+    marker: {
+        backgroundColor: 'blue'
+    }
+};
+```
+
 ## Server-side rendering
 
 To perform server-side rendering, make a call to `StyleSheetServer.renderStatic`, which takes a callback. Do your rendering inside of the callback and return the generated HTML. All of the calls to `css()` inside of the callback will be collected and the generated css as well as the generated HTML will be returned.
--- a/src/inject.js
+++ b/src/inject.js
@@ -1,7 +1,7 @@
 import asap from 'asap';
 
 import {generateCSS} from './generate';
-import {hashObject} from './util';
+import {flattenDeep, hashObject} from './util';
 
 // The current <style> tag we are inserting into, or null if we haven't
 // inserted anything yet. We could find this each time using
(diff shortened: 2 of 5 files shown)


jlfwong:
@ide Thanks for the patch, with tests no less!

@xymostech Thoughts? I'm happy with doing something like this, though I'd prefer it be recursive as well.

xymostech:
I'm pretty happy about merging this! I think that the marginal increase in API complexity is worth it, especially since this is what React Native does.

I agree though that this should almost certainly be recursive, otherwise people will be confused about when nesting lists is acceptable.

ide:
@jlfwong @xymostech thanks! I'll make the flattening be recursive -- are you ok with an extra dependency (https://www.npmjs.com/package/lodash.flattendeep) or would you prefer flattenDeep to live under utils.js?

xymostech:
I'd be fine with the dependency, but I'm never sure how small the individual lodash packages are. Can you add the dependency and see how big the built file gets?

jlfwong:
I'd prefer no dependency. We don't need to cover as many cases as the lodash package does. It seems pretty expansive: https://github.com/lodash/lodash/blob/4.4.0-npm-packages/lodash.flattendeep/index.js. It deals with things like `arguments` objects which I don't feel a compelling reason to support. 

Does React Native support non-Array types like iterators or enumerable things?

ide:
React Native supports only arrays I think. Supporting other iterables like Sets could be nice though, iterall makes that easy and is fairly lightweight, it's 600KB minified and gzipped. https://github.com/leebyron/iterall

Do you prefer supporting iterables or just arrays?

jlfwong:
Here's the relevant React Native source (I think). https://github.com/facebook/react-native/blob/master/Libraries/StyleSheet/flattenStyle.js

Arrays only, please!

jlfwong:
You'll also need to sign this in case you haven't already https://docs.google.com/a/figma.com/forms/d/e/1FAIpQLSdyXYrc8ogVoA46J9KXyIj5nKlZzNkOnQG-4A1R7X_BWGTShQ/viewform

ide:
I signed the CLA and added `flattenDeep` plus a couple of tests.

xymostech:
Awesome. I'm happy with this if you are, @jlfwong!

jlfwong:
Hooray! Last thing is to add a section to the README documenting this, and then it's good to go!

I think it should go right below "Conditionally Applying Styles"

ide:
Added a README section (preview here https://github.com/ide/aphrodite/tree/css-arrays#combining-styles). I used the example of combining styles from props since that answers some questions that other people have asked in a couple of older issues.

kevinbarabash:
@ide thanks for signing our CLA.

jlfwong:
Sweeet - let's get this landed!

PR #157, opened 2016-10-10
really try insertRule

### LINKED ISSUE #147: Styles flicker in.
Hey there,

Loving aphrodite so far. I keep noticing that when my components render on the screen there is a small flicker. The component is first unstyled then styled. I am not using SSR. 

Is there a way to preload css functions? Or am I just doing something wrong?

### LINKED ISSUE #155: Support pseudo elements under pseudo styles
I was whippin up a PR for #147 when I noticed that the following code is not supported:
```
const styles = {
  input: {
    ':focus': {
      '::-webkit-input-placeholder': {
        color: 'blue'
      }
    }
  }
}
```
[Example](https://output.jsbin.com/qoseye?code=%7B%0A%20%20input%3A%20%7B%0A%20%20%20%20%27%3Afocus%27%3A%20%7B%0A%20%20%20%20%20%20%27%3A%3A-webkit-input-placeholder%27%3A%20%7B%0A%20%20%20%20%20%20%20%20color%3A%20%27blue%27%0A%20%20%20%20%20%20%7D%0A%20%20%20%20%7D%0A%20%20%7D%0A%7D)

### LINKED ISSUE #156: Support inlined fonts
Per #82, `src` is used as the unique insertion key.
This is bad news bears if someone base64 inlines their fonts, since they'll be sending down a rehydrations object with keys that are 10KB each.

maybe something like this?

```
    if (Array.isArray(val)) {
      return val.map(fontFamily).join(",");
    } else if (val && typeof val === "object") {
      const {fontFamily, fontStyle, fontWeight} = val;
      const key = `${fontFamily}-${fontWeight || 400}${fontStyle}`;
      injectStyleOnce(key, "@font-face", [val], false);
      return fontFamily;
    } else {
      return val;
    }
  },
```

mattkrick:
Features:
- ~~Fix #156 by generating short, unique font names~~ See #158
- Fix #155 by roughly treating pseudos like media queries
- Fix #147 by using insertRule
- ~~Remove `px` from `0` _because every byte counts_™~~ See #159
- adds 4-space indent to eslint

I'm betting this is gonna be a hard sell, so here's the pitch:
- Touching a style tag after creation is bad. Really bad. After changing it, the DOM must parse the text, then evaluate the rules. ALL the rules. This isn't cheap, and it grows in O(n) time as more styles get added. We could solve this by creating multiple tags, but having 4000 style tags is tough to debug. Plus, if you exceed that, IE10 yells at you. We can kick the can down the road by sticking fonts in their own tag, but then you get a flicker on initial load & for every font you async load. Also, if a tag gets large enough, it'll cause a flicker regardless of a font being in there or not.
- This allows for styles to fail loudly. I think bugs like #155 went uncaught largely because browsers parse style tags and toss out anything that doesn't make sense. The tossing stuff out is a silent fail. By using an API that doesn't fail silently, we can use it down the road when we create a dev/production build. Could you imagine? You could run QA _without needing to know what the page is supposed to look like_ because it just logs an error if a style didn't load. I did that with my own app while writing this. It's a treat 👍 
-  This opens the door to removing `inline-styles-prefixer` from the client build, saving ~3KB
- This opens the door to removing `asap` since injections are so insanely cheap. This would solve #76 in the best way possible.

How it works:
The buffer used to store an big glob of text comprised of many rules. Now, the buffer stores an array of rules + for each rule a boolean called `isDangerous`. If a rule is dangerous, we use a try/catch block, which means V8 will bail on that function and never try to optimize it. That's fine because the calling function is externalized and the only things that trigger a dangerous flag are the vendor-proprietary things that they don't list in their computedStyles: `::-moz-placeholder, ::-moz-focus-inner` or already prefixed globals that user is manually trying to inject. Otherwise, we inject with confidence. 

We could whitelist these things, but by not using a whitelist, we save a little space in the payload. More importantly, we don't have to update the package whenever a browser vendor adds something. Also worth noting that the call is memoized so we only make 1 call to `window.getComputedStyles` and that array learns browser prefixes, so aphrodite gets (negligibly) faster the more you use it.

By avoiding `prefixAll`, we only generate rules that the client can understand. This is great because it is faster (less to generate = less to inject), gets rid of prefix problems (#100), and opens up the door to more helpful debugging.

Happy to answer questions. I know it's a big change, but for large apps, it's a necessary one. I know there are a few other folks like @abhiaiyer91 with large apps who have been struggling with the same performance bottlenecks, so I wanted to provide an option to stay competitive with JSS.


xymostech:
Hey @mattkrick!

Thanks for putting the effort into making these changes. Before trying to polish up the code, I have some broad feedback that I think will help us get this landed:

- There are a lot of changes that you're making here, some which I think we want and some which will require some more thought. Putting them all into one pull request is going to make it hard to talk about the changes that are actually important here. In particular, I think a pull request that _only_ adds the use of `insertRule` would provide much more useful discussion. Making changes to the font family keys or rewriting our stringify function at the same time is just going to make it harder for us to land.
- The changes you made are failing a lot of tests (as you can see [over at travis](https://travis-ci.org/Khan/aphrodite/builds/166566285)). Fixing up those tests and getting back to 100% code coverage will probably help us understand changes to the API better, and help us make sure these changes are reasonable.
- It would be great if you followed our style guide (in particular, 4 space indents and 80 character line length limits)

So, if you could strip this down to the smallest diff that still demostrates what using `insertRule` looks like, and getting the tests to pass once you do that, that would be great!

mattkrick:
@xymostech oh wow, i didn't actually think this would get considered for master :yum:. Yeah, it'll require a _lot_ of rewriting of tests & jsdocs. Happy to do it if it's actually gonna get merged, but writing tests for rejected PRs is no good. Any feedback you can give me before then? 

PS, any chance KA might adopt an `.eslintrc` or offer up an `npm run lint` command?

xymostech:
@mattkrick I just landed the pull request to add an eslintrc, `npm run lint` should work now if you rebase :)

xymostech:
@mattkrick As for advice, if this fixes the problems in #147 then I'm really interested in it. So, test to make sure it acutally solves that problem before putting in the effort?

mattkrick:
@xymostech it definitely solves it locally on firefox, chrome, and opera (we could even remove the 'o' prefix check, that hasn't been used by opera in...years?). Haven't tried IE since I don't have a windows box, but I'm pretty sure the `getComputedStyles` camelcase prefix is correct. 
I'll update the staging server that you see in #147 so you can verify that the flicker is gone & then get some tests written up

mattkrick:
Alrighty, submitted 2 PRs that take care of the extra stuff. After those are merged, this PR will only have stuff relevant to #147. #155 can't be broken out since that bug prevents it from working.

mattkrick:
@xymostech what's your secret for getting 100% branch coverage? 🙈

xymostech:
Looks like you don't have any lines left to cover, but the command line output doesn't help with the other coverages. If you run

```
./node_modules/.bin/nyc --reporter=lcov npm run tests
```

it'll generate a much more usable HTML format for you to browse the coverage. In your case, the things I'm finding that aren't covered are:

generate.js:75
image

inject.js:109
image

inject.js:37 (you're always passing the "if" check here, and there's no else block. Maybe just get rid of this if)
image

util.js:202
image

util.js:246
image

It would be cool if this was auto-generated and hosted somewhere if the coverage fails...

mattkrick:
@xymostech ready for your 👀 . 

I added the 4-space indent rule so i didn't have to manually fix up my stuff, it found some pre-existing stuff that wasn't compliant, hence the touches to webpack configs.

chendo:
@mattkrick hopefully I have time to try this branch again today.

jlfwong:
> We can kick the can down the road by sticking fonts in their own tag, but then you get a flicker on initial load & for every font you async load. 

I'm a little confused by this. Why would there be a flicker on initial load? We could make one tag per font. Presumably you won't be using enough fonts to hit the IE10 limit.

> Also, if a tag gets large enough, it'll cause a flicker regardless of a font being in there or not.

Is this an assumption or is there proof? I suspect the browser might lag a little, but it won't show you invalid results. I'm not certain either way though.

mattkrick:
@jlfwong The initial flicker would be caused by a switch from the fallback font to the downloaded font (assuming it doesn't get flushed immediately).If you can flush immediately, you shouldn't get a flash.

I'd also be very interested to hear from the folks who still see a flash after removing the fonts if this branch makes the flash go away entirely. TBH I thought a font that was inlined in base64 would be loaded in sync & wouldn't cause a flash, but it's still treated as a (cached) network request. You'd think the browser would create a checksum of the fetched file & compare it with the current...

src/util.js
```diff
@@ -197,3 +191,58 @@ export const importantify = (string) =>
...
+    if (validatedProp) {
+        return validatedProp;
+    }
+    const capitalProp = property[0].toUpperCase() + property.substr(1);
+    for (let i = 0; i < stylePrefixes.length; i++) {
+        const prefix = stylePrefixes[i];
+        const prefixedProperty = `${prefix}${capitalProp}`;
+        if (validProperties[prefixedProperty]) {
+      // learn which styles the browser likes
+            validProperties[validatedProp] = prefixedProperty;
```
wereHamster: `validatedProp` is falsy here (most likely `undefined`). Don't you want `validProperties[property] == ...` instead?

abhiaiyer91:
@mattkrick testing this branch on our app now

chendo:
@mattkrick I've tried using the newest commit of this branch (https://github.com/mattkrick/aphrodite/tarball/d131a02923e6daa62df223e77b791eecaca9e067) and ran `npm run build:main`, but it appears to break compatibility with `react-with-styles` and `react-with-styles-interface-aphrodite` as `StyleSheet` doesn't appear to be defined anymore.

screenshot 2016-10-17 13 39 41
screenshot 2016-10-17 13 40 11

src/util.js
mattkrick: Let's assume you have a rule like `boxReflect`.
It first checks to see if the browser can handle that by looking up `boxReflect` (Line 213).
If that doesn't work, then it tries all the prefixes. `MozBoxReflect, webkitBoxReflect`, ah, found it!
We know that the browser doesn't support `boxReflect` but it does support `webkitBoxReflect`.
So now we add that to the validatedProperties table: `{boxReflect: 'webkitBoxReflect'}`.
So the next time you look up `boxReflect` (Line 213) you grab the prefixed version immediately without having to set up the for loop.  
The time savings is minimal, but aphrodite's target is IE10, so we can't use a `Set`, so we might as well make use of the object.

mattkrick:
@chendo i don't think you're referencing the branched version correctly. This PR doesn't do anything to the exports. For your convenience, try installing `aphrodite-local-styles` via npm and replace all references of `'aphrodite'` to `'aphrodite-local-styles'`. Note: I'll remove that package from npm if this gets merged.

src/util.js
wereHamster: I understand that. But following the control flow of the code, at this line (223), the value of `validatedProp` is falsy (an empty string (`""`), zero (0), `false`, `undefined` or `null`). Doing simple substitution, we arrive at either one of the following lines of code:

```
validProperties[undefined] = prefixedProperty;
validProperties[""] = prefixedProperty;
validProperties[0] = prefixedProperty;
validProperties[false] = prefixedProperty;
```

Is that really what you intended to do?

chendo:
@mattkrick got it to work! Significantly decreased the amount of unstyled flicker we see, but we still see it occasionally in some scenarios. Thanks!

mattkrick:
@chendo in what scenario do you see a flicker still?

abhiaiyer91:
@mattkrick EARLY Reads. `aphrodite-local-styles` work. Reduces flicker. Only reason flicker remains is I need to change more imports to local styles. I think we may go forward with this in the short term. our users are getting a terrible experience now

mattkrick:
@abhiaiyer91 you mean you have both aphrodite & my fork of it installed?

yanivtal:
This is exciting! I guess it'll have to be rebased ontop of the latest extension changes. I noticed Aphrodite was just bumped to 1.0.0. Will that slow down merging this feature?

yanivtal:
Cool, no problem. Just making sure since it seems like this'll help a lot with perf. Just out of curiosity, is there any discussion about chunking across style tags for more perf? Seems we still have a big hit on every insert rule unless I'm missing something. @mattkrick says the multiple tags are harder to debug but I'm not familiar with that problem.

yanivtal:
@mattkrick was this just waiting for Khan approval?

mattkrick:
Yup

abhiaiyer91:
@mattkrick @xymostech Are we going to get this into master anytime soon?

bernardoatworkpop:
Hey @mattkrick @xymostech, seems like the thread hasn't been updated in a while. Is this fix still planned to happen? thx

mattkrick:
¯\_(ツ)_/¯ At this point, I doubt it. FWIW, I use aphrodite-local-styles in production and it works great. Since I made this PR, Aphrodite now supports mixing global and local styles. I think that's dangerous, so I have no desire to keep my fork up to date with master when my version is more performant and disallows anti patterns. In the future, I'll shrink the client payload for mine and insert rules in sync so you don't have to use setTimeout hacks. If you'd like to take over getting this merged, feel free, but I'll close for now as it looks like my goals are no longer aligned with the project.

PR #158, opened 2016-10-11
Dont use src for font injection key fix #156

### LINKED ISSUE #156: Support inlined fonts

mattkrick:


jaredly:
what's the error this is fixing?

mattkrick:
fonts can be base64 inlined

PR #159, opened 2016-10-11
dont add 'px' suffix for 0 values

mattkrick:


jaredly:
this looks useful! However, it changed the style rather a lot -- could you keep the original setup and just add a `|| value === 0` to the one if condition?

PR #162, opened 2016-10-14
add FlowType annotations

jokeyrhyme:
-   I prefer my code to remain valid and executable ECMAScript syntax, so I've added the FlowType annotations using the optional comment style: https://flowtype.org/blog/2015/02/20/Flow-Comments.html
  -   I can switch this to real FlowType syntax if you like, as you're already using Babel anyway
-   I copied a few basic FlowType settings from yarn: https://github.com/yarnpkg/yarn/blob/v0.15.1/.flowconfig
- while mostly arbitrary, some of the longer function signatures were broken into multiple lines, and some of the shorter function signatures remain on a single line
- the one suppression I had to put in (`// $FlowFixMe: ...`) was because the built-in declaration of `querySelector()` is that it returns an `HTMLElement`, and FlowType strictly does not allow us to downcast to the `HTMLStyleElement` class as required here

### DIFF
--- a/src/exports.js
+++ b/src/exports.js
@@ -1,3 +1,4 @@
+/* @flow */
 import {mapObj, hashObject} from './util';
 import {
     injectAndGetClassName,
@@ -5,8 +6,19 @@ import {
     addRenderedClassNames, getRenderedClassNames,
 } from './inject';
 
+/* ::
+import type { SelectorHandler } from './generate.js';
+export type SheetDefinition = { [id:string]: any };
+export type SheetDefinitions = SheetDefinition | SheetDefinition[];
+type RenderFunction = () => string;
+type Extension = {
+    selectorHandler: SelectorHandler
+};
+export type MaybeSheetDefinition = SheetDefinition | false | null | void
+*/
+
 const StyleSheet = {
-    create(sheetDefinition) {
+    create(sheetDefinition /* : SheetDefinition */) {
         return mapObj(sheetDefinition, ([key, val]) => {
             return [key, {
                 // TODO(emily): Make a 'production' mode which doesn't prepend
@@ -17,7 +29,7 @@ const StyleSheet = {
         });
     },
 
-    rehydrate(renderedClassNames=[]) {
+    rehydrate(renderedClassNames /* : string[] */ =[]) {
         addRenderedClassNames(renderedClassNames);
     },
 };
@@ -26,7 +38,7 @@ const StyleSheet = {
  * Utilities for using Aphrodite server-side.
  */
 const StyleSheetServer = {
-    renderStatic(renderFunc) {
+    renderStatic(renderFunc /* : RenderFunction */) {
         reset();
         startBuffering();
         const html = renderFunc();
@@ -76,7 +88,10 @@ const StyleSheetTestUtils = {
  * Generate the Aphrodite API exports, with given `selectorHandlers` and
  * `useImportant` state.
  */
-const makeExports = (useImportant, selectorHandlers) => {
+const makeExports = (
+    useImportant /* : boolean */,
+    selectorHandlers /* : SelectorHandler[] */
+) => {
     return {
         StyleSheet: {
             ...StyleSheet,
@@ -97,7 +112,7 @@ const makeExports = (useImportant, selectorHandlers) => {
              * @returns {Object} An object containing the exports of the new
              *     instance of Aphrodite.
              */
-            extend(extensions) {
+            extend(extensions /* : Extension[] */) {
                 const extensionSelectorHandlers = extensions
                     // Pull out extensions with a selectorHandler property
                     .map(extension => extension.selectorHandler)
@@ -115,7 +130,7 @@ const makeExports = (useImportant, selectorHandlers) => {
         StyleSheetServer,
         StyleSheetTestUtils,
 
-        css(...styleDefinitions) {
+        css(...styleDefinitions /* : MaybeSheetDefinition[] */) {
             return injectAndGetClassName(
                 useImportant, styleDefinitions, selectorHandlers);
         },
--- a/.flowconfig
+++ b/.flowconfig
@@ -2,8 +2,9 @@
 <PROJECT_ROOT>/.nyc_output/.*
 <PROJECT_ROOT>/coverage/.*
 <PROJECT_ROOT>/dist/.*
-<PROJECT_ROOT>/node_modules/babel.*
-<PROJECT_ROOT>/node_modules/y18n/test/.*
+.*/node_modules/babel.*
+.*/node_modules/fbjs/.*
+.*/node_modules/y18n/test/.*
 
 [include]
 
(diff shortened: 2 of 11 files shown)


wereHamster:
You can avoid the `$FlowFixMe` when you add an instanceof check. It would be safer anyways.

jokeyrhyme:
@wereHamster yeah, I need to double-check how that part works anyway. That's actually part of an old Internet Explorer compatibility path, which isn't part of the DOM that FlowType already knows about.

wereHamster:
I mean, you can do something like:

```javascript
if (styleTag instanceof HTMLStyleElement && styleTag.styleSheet) {
  styleTag.styleSheet.cssText += cssContents;
} else {
  styleTag.appendChild(document.createTextNode(cssContents));
}
```

jokeyrhyme:
@wereHamster what I mean is that W3C `HTMLStyleElement` objects do not even have a "styleSheet" property. That's only in legacy Internet Explorer. But yeah, I'll play with `instanceof` and see if there's an alternative to the FlowType suppression.

wereHamster:
Oh, then I don't know how to do it without the `$FlowFixMe` :(

jokeyrhyme:
@wereHamster I had another attempt and couldn't come up with something nicer than the FlowType suppression. We do have ESLint suppressions in a few places, so maybe this isn't so bad?

jokeyrhyme:
Bah, just found https://github.com/flowtype/flow-typed/blob/master/definitions/npm/aphrodite_v0.5.x/flow_v0.28.x-/aphrodite_v0.5.x.js

xymostech:
@jokeyrhyme Sorry I haven't reviewed this yet! This looks really valuable, in terms of internal documentation and bug catching, even if there is external flow types available.

I just landed a big patch that caused some conflicts, if you update for that I'd be happy to take a look again! (I'm happy with having `$FlowFixMe`s for now by the way. Better to get the changes in there than try to fix everything now).

Also could you sign our cla? https://docs.google.com/a/khanacademy.org/forms/d/e/1FAIpQLSdyXYrc8ogVoA46J9KXyIj5nKlZzNkOnQG-4A1R7X_BWGTShQ/viewform Thanks!

jokeyrhyme:
I've signed the CLA as part of a few other PRs that have been merged.

I'll take a look at the conflicts over the next few days.

jokeyrhyme:
Also, if you could assign this PR to me that would be great. Helps me keep track of things. :)

jokeyrhyme:
@xymostech okay, I've resolved the conflicts. Nice bit of refactoring, I see. :)

jokeyrhyme:
Figured it might be worth giving a heads up to @ajhyndman @thejameskyle @wbinnssmith @jeffmo (they worked on the flow-typed definition)

ajhyndman:
Awesome!  I'm really happy to see this proposal.  I expect that the flow types are much more likely to stay up-to-date with a library, when incorporated in the code.

👍

src/exports.js
```diff
@@ -115,7 +129,7 @@ const makeExports = (useImportant, selectorHandlers) => {
         StyleSheetServer,
         StyleSheetTestUtils,
 
-        css(...styleDefinitions) {
+        css(...styleDefinitions /* : SheetDefinition[] */) {
```
ajhyndman: Is the conditional style API still supported?

https://github.com/Khan/aphrodite#conditionally-applying-styles

It might be worth widening the definition to accept at least the boolean `false` here, as was done in the flow-typed definition.

https://github.com/flowtype/flow-typed/commit/c47ccdab811f52e1afb68492689330421030b344

jokeyrhyme:
@ajhyndman good find. :) I've added a "MaybeSheetDefinition" type that is either a SheetDefinition or falsey.

xymostech:
Hi @jokeyrhyme. Sorry this has taken me so so long to review. The code in here looks wonderful. And it looks like the tests are passing on travis, so I'm happy to merge this. I'm a little confused though: I cloned this myself to check out what the flow output looks like, and I'm getting a bunch of errors about "required module not found":

```
% ./node_modules/.bin/flow check
examples/node_modules/react/node_modules/fbjs/lib/Deferred.js.flow:14
 14: var Promise = require('Promise');
                   ^^^^^^^^^^^^^^^^^^ Promise. Required module not found

examples/node_modules/react/node_modules/fbjs/lib/PromiseMap.js.flow:15
 15: var Deferred = require('Deferred');
                    ^^^^^^^^^^^^^^^^^^^ Deferred. Required module not found

examples/node_modules/react/node_modules/fbjs/lib/PromiseMap.js.flow:17
 17: var invariant = require('invariant');
                     ^^^^^^^^^^^^^^^^^^^^ invariant. Required module not found

examples/node_modules/react/node_modules/fbjs/lib/PromiseMap.js.flow:19
 19: import type Promise from 'Promise';
                              ^^^^^^^^^ Promise. Required module not found

examples/node_modules/react/node_modules/fbjs/lib/fetchWithRetries.js.flow:16
 16: var ExecutionEnvironment = require('ExecutionEnvironment');
                                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ ExecutionEnvironment. Required module not found

examples/node_modules/react/node_modules/fbjs/lib/fetchWithRetries.js.flow:17
 17: var Promise = require('Promise');
                   ^^^^^^^^^^^^^^^^^^ Promise. Required module not found

examples/node_modules/react/node_modules/fbjs/lib/fetchWithRetries.js.flow:19
 19: var sprintf = require('sprintf');
                   ^^^^^^^^^^^^^^^^^^ sprintf. Required module not found

examples/node_modules/react/node_modules/fbjs/lib/fetchWithRetries.js.flow:20
 20: var fetch = require('fetch');
                 ^^^^^^^^^^^^^^^^ fetch. Required module not found

examples/node_modules/react/node_modules/fbjs/lib/fetchWithRetries.js.flow:21
 21: var warning = require('warning');
                   ^^^^^^^^^^^^^^^^^^ warning. Required module not found

examples/node_modules/react/node_modules/fbjs/lib/resolveImmediate.js.flow:13
 13: var Promise = require('Promise');
                   ^^^^^^^^^^^^^^^^^^ Promise. Required module not found


Found 10 errors
```

I haven't really used flow before, do you know what could be causing this?

jokeyrhyme:
@xymostech my examples/node_modules does not exit

So did this `cd examples; npm install`, and then I could reproduce the errors

For some of Facebook's JavaScript code, they've [customised the way `require()` works](https://facebook.github.io/react/contributing/codebase-overview.html#custom-module-system)

This means linters that check to see whether module references are satisfied are likely to be confused by Facebook's code

I've pushed some `[ignore]` statements for FlowType's configuration that should prevent these files from being scanned

xymostech:
Cool! Looks really great to me! I have a question about how this works for people importing Aphrodite code: Do people (who are using flow) automatically get to see the annotations that are added here? Or do we need to export them or something?

xymostech:
[clabot:check]

jokeyrhyme:
For people using Aphrodite and FlowType in their own projects, the types of arguments, etc will be checked where they use the Aphrodite APIs, as per the types defined within.

However, only some of the types are properly exported for now, which means they won't be available for downstream consumers to `extend` or otherwise use within their own code. If this is important, we can have a new PR to properly export all types.

It'll be interesting to see feedback from users in this regard.

PR #164, opened 2016-10-25
Update Other solutions in README

DjebbZ:
One of them is officially abandonned, the other looks abandonned too. Glamor is in active development on the other hand.

### DIFF
--- a/README.md
+++ b/README.md
@@ -486,9 +486,8 @@ It isn't determinate whether divs will be red or blue.
 
 # Other solutions
 
-- [js-next/react-style](https://github.com/js-next/react-style)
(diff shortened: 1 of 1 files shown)


xymostech:
Hi @DjebbZ! Thanks for updating this. Could you add `jss` also? I think that's the other library that's in about the same space as us. Other than that, looks great!

jcperez-ch:
I +1 JSS, it comes really strong, and the author recently made a comparison between `JSS` and `aphrodite` also, it has a super set named [`aphrodisiac`](https://github.com/cssinjs/aphrodisiac) which brings the cool things of aphrodite on the top of JSS core.

xymostech:
@DjebbZ I'm going to close this for now. Feel free to re-open if you update this.

PR #187, opened 2017-01-23
Fix generateCSS tests

borkxs:
Make generateCSS tests not depend on sort behavior of inline-style-prefixer.

A simpler solution would be to just change the expect string for these tests but they're actually depending on the behavior of [ineline-style-prefixer's sort method](https://github.com/rofrischmann/inline-style-prefixer/blob/master/modules/utils/sortPrefixedStyle.js)

### DIFF
--- a/tests/generate_test.js
+++ b/tests/generate_test.js
@@ -32,11 +32,12 @@ describe('generateCSSRuleset', () => {
     });
 
     it('prefixes vendor props with a dash', () => {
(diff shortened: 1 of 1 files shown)


borkxs:
[clabot:check]

xymostech:
Oops, I didn't see your PR for this and pushed up a fix already! :P Thanks for taking the initiative to fix, though!

PR #188, opened 2017-01-24
Fix to prevent arrays being mutated into objects

lukevenn:
When recursively merging it will currently convert an array into an object which, when merging two styles with `fontFamily` properties, cause the generated style to set it as `"undefined"`.

e.g. `font-family: "undefined"`

### DIFF
--- a/src/util.js
+++ b/src/util.js
@@ -37,6 +37,10 @@ const MS_RE = /^ms-/;
 const kebabify = (string /* : string */) /* : string */ => string.replace(UPPERCASE_RE, '-$1').toLowerCase();
 export const kebabifyStyleName = (string /* : string */) /* : string */ => kebabify(string).replace(MS_RE, '-ms-');
 
+const isNotObject = (
+  x/* : ObjectMap | any */
+) /* : boolean */ => typeof x !== 'object' || Array.isArray(x) || x === null;
+
 export const recursiveMerge = (
     a /* : ObjectMap | any */,
     b /* : ObjectMap */
@@ -44,7 +48,7 @@ export const recursiveMerge = (
     // TODO(jlfwong): Handle malformed input where a and b are not the same
     // type.
 
-    if (typeof a !== 'object' || Array.isArray(b)) {
+    if (isNotObject(a) || isNotObject(b)) {
         return b;
     }
 
--- a/src/util.js
+++ b/src/util.js
@@ -44,7 +44,7 @@ export const recursiveMerge = (
     // TODO(jlfwong): Handle malformed input where a and b are not the same
     // type.
 
-    if (typeof a !== 'object') {
+    if (typeof a !== 'object' || Array.isArray(b)) {
         return b;
     }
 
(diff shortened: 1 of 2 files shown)


lukevenn:
`[clabot:check]`

src/util.js
```diff
@@ -44,7 +44,7 @@ export const recursiveMerge = (
     // TODO(jlfwong): Handle malformed input where a and b are not the same
     // type.
 
-    if (typeof a !== 'object') {
+    if (typeof a !== 'object' || Array.isArray(b)) {
```
xymostech: Do we maybe want to check if `a` is an array as well? We want merging `{a: [1]}` and `{a: 2}` to do something reasonable here.

tests/util_test.js
```diff
@@ -34,5 +34,16 @@ describe('Utils', () => {
                     b: 2,
                 });
         });
+        it('does not mutate Arrays into Objects', () => {
```
xymostech: The arrays aren't being mutated, it's more that they're being replaced by objects, so I'm not sure I agree with this name.

For future readers though (who don't have context about what was happening in this bug) I'd prefer if we didn't call this "it doesn't..." and instead say "it does..." like "it replaces arrays instead of trying to merge them." How does that sound?

src/util.js
lukevenn: I did think about merging arrays but in the use case where I was seeing the bug (with fontFamily) I don't think that you want to merge because it would cause an unexpected result. As an example, suppose you have a base style that has a fontFamily array and a second style that you want to merge that overrides the fontFamily:
```
const base = {...
  fontFamily: ['Arial', 'Helvetica', 'sans-serif'],
...}
const override = {...
  fontFamily: ['Times New Roman', 'serif'],
...}
```
The you'd expect the merged style to have a `fontFamily` value that matched `override` but if you merged you'd end up with:
```
fontFamily: ['Arial', 'Helvetica', 'sans-serif', 'Times New Roman', 'serif']
```
which would cause unexpected styling.

tests/util_test.js
lukevenn: Good point. I will try to come up with a better title.

src/util.js
lukevenn: @xymostech Sorry, just re-read your comment and realised I completely missed the point. So do you think the code should be something like:
```
if (typeof a !== 'object' || Array.isArray(a) || Array.isArray(b)) {
```
Actually I think we probably need to take it further
```
if (typeof a !== typeof b && typeof a !== 'object'
  || Array.isArray(a)
  || Array.isArray(b)
  || a === null
  || b === null
) {
```
as this would then cover off the situation where we want to merge `{a:  {b: 2}}` and `{a: 2}` and when we're merging `{a: 1}` with `{a: null}` ...although it doesn't look pretty.

Let me know your thoughts

lukevenn:
Hi @xymostech, I've just added in a simple object check that will hopefully cover merging of different types. Essentially, I've taken the view that if either `a` or `b` aren't objects then we can't merge anyway so we should just take `b`.

Issue #189, opened 2017-01-31
Question about rehydrate


lazopm:
Is there any documentation on what exactly rehydrate does?

One of my pre-rendered react components is working well without using rehydrate at all. 

I'm throwing the generated css in a <style> tag next to the react-root div. Source looks like this:
```
<style data-aphrodite>...</style>
<div react-root>...</div>
```

What do I do if I have several react components in a page that get pre-rendered separately? The example in the readme is for a single root component.

Thank you!

xymostech:
Rehydrating lets Aphrodite know which styles have already been inserted into the page. If you don't rehydrate, Aphrodite might add duplicate styles to the page, which might be bad for performance.

If you have several react components, you can call `StyleSheet.rehydrate` once for each of your components with the list of class names that you get from server-side rendering. For instance:

```js
// on the server
var {html1, css1} = StyleSheetServer.renderStatic(() => {
    return ReactDOMServer.renderToString(<Component1 />);
});
var {html2, css2} = StyleSheetServer.renderStatic(() => {
    return ReactDOMServer.renderToString(<Component2 />);
});
```

Then you inject:
```html
<style data-aphrodite>${css1.content}${css2.content}</style>
```

And on the client, you rehydrate twice:
```js
StyleSheet.rehydrate(css1.renderedClassNames);
StyleSheet.rehydrate(css2.renderedClassNames);
```

Does that make sense?

(I'm going to close this issue because it's not related to an actual bug in the project. Feel free to continue the discussion here though!)

lazopm:
Makes perfect sense, thank you!

lazopm:
Hey @xymostech, I was thinking, is there a reason why rehydrate can't be automatic?
Aphrodite could rehydrate all the classes in <style> tags that have `data-aphrodite`.
If analyzing the contents of the style tag is too slow, maybe `renderStatic` could generate a css comment that has a list of all the classes in that particular css block.
What do you think? Would there be a downside to that?

PR #190, opened 2017-01-31
Add description of rehydration to README.md

lazopm:
Added @xymostech's description of rehydration from #189 to the readme.
I think it will be helpful for new users.


lazopm:
[clabot:check]

xymostech:
Looks mostly good to me! I think you should remove "which might be bad for performance." because that's more wild speculation than actual fact. :)

xymostech:
Thanks!

PR #192, opened 2017-02-03
Little update for extend

tuchk4:
Little README update - more clear usage of `extends` method.


tuchk4:
[clabot:check]

xymostech:
Ah yes! Thank you for that!

PR #200, opened 2017-03-03
Allow specifying styles as `Map`s to guarantee ordering

### LINKED ISSUE #199: Order of properties can change in different environments
I am noticing a difference in how the style properties are ordered between a development environment and a production environment. Essentially, I have a couple of styles that look like this:

```js
foo: {
  margin: 0,
},

bar: {
  marginRight: 16,
},
```

I am applying them to my element like this `css(foo, isBar && bar)`.

In my development environment (a story in react storybook), the component is styled as expected, with 0 margin all around except for the 16px of margin on the right. However, in production, my 16px right margin is missing. Looking at the generated styles, I can see that all of my properties are there, but the order is different.

In storybook:

```css
.an_aphrodite_selector {
  margin: 0 !important;
  margin-right: 16px !important;
}
```

In production:

```css
.an_aphrodite_selector {
  margin-right: 16px !important;
  margin: 0 !important;
}
```

I also have some more styles than just margin, but I am just mentioning these because it highlights the specific problem. Here's the specific CSS that is generated in development (whitespace added for readability):

```css
.text_1aunhvg-o_O-tab_8ndy7u-o_O-tab_micro_17dr3ug-o_O-text_muted_td6kew-o_O-text_spacedTabs_1fn1i59-o_O-tab_selected_6lr1xw{
	-webkit-transition-timing-function:ease-out !important;
	-webkit-transition-property:color !important;
	-webkit-transition-duration:300ms !important;
	letter-spacing:0.4px !important;
	padding-top:12px !important;
	padding-bottom:12px !important;
	color:#008489 !important;
	margin:0px !important;
	background-color:transparent !important;
	border-width:0px !important;
	border-bottom:2px solid #008489 !important;
	font-size:12px !important;
	cursor:pointer !important;
	margin-right:16px !important;
	padding-left:0px !important;
	padding-right:0px !important;
	font-weight:700 !important;
	transition-property:color !important;
	transition-duration:300ms !important;
	transition-timing-function:ease-out !important;
	font-family:Circular,-apple-system,BlinkMacSystemFont,Roboto,Helvetica Neue,sans-serif !important;
	line-height:16px !important;
	bottom:-1px !important;
}
```

and in production:

```css
.text_1aunhvg-o_O-tab_8ndy7u-o_O-tab_micro_ltghyi-o_O-text_muted_td6kew-o_O-text_spacedTabs_1fn1i59-o_O-tab_selected_6lr1xw{
	padding-left:0px !important;
	bottom:-1px !important;
	font-weight:700 !important;
	border-width:0px !important;
	border-bottom:2px solid #008489 !important;
	cursor:pointer !important;
	padding-top:12px !important;
	font-size:12px !important;
	letter-spacing:0.4px !important;
	font-family:Circular,-apple-system,BlinkMacSystemFont,Roboto,Helvetica Neue,sans-serif !important;
	margin-right:16px !important;
	padding-right:0px !important;
	padding-bottom:18px !important;
	line-height:16px !important;
	color:#008489 !important;
	margin:0px !important;
	background-color:transparent !important;
	transition-property:color !important;
	transition-duration:300ms !important;
	transition-timing-function:ease-out !important;
	-webkit-transition-property:color !important;
	-webkit-transition-duration:300ms !important;
	-webkit-transition-timing-function:ease-out !important;
}
```

Any ideas what might be happening here or where to look? It seems that there might be some code that expects the order of things to be stable when that is not the case. Or perhaps we are transforming (Babel) or polyfilling something slightly differently in these two environments? 

Aphrodite v0.5.0

xymostech:
Summary: Key ordering in objects can be different in different environments.
Sometimes, this causes problems, like where styles generated on the server are
not in the same order as the same styles generated on the client. This
manifests in problems such as #199

This change lets users manually fix instances where the ordering of elements
changes by specifying their styles in an ES6 `Map`, which has defined value
ordering.

In order to accomplish this, an `OrderedElements` class was created, which is
sorta like a `Map` but can only store string keys and lacks most of the
features. Internally, `Map`s and objects are converted into this and then these
are merged together to preserve the ordering.

Fixes #199

Test Plan:
 - `npm test`

@lencioni @ljharb

### DIFF
--- a/README.md
+++ b/README.md
@@ -429,6 +429,82 @@ then we end up with the opposite effect, with `foo` overriding `bar`! The way to
 }
 ```
 
+## Object key ordering
+
+When styles are specified in Aphrodite, the order that they appear in the
+actual stylesheet depends on the order that keys are retrieved from the
+objects. This ordering is determined by the JavaScript engine that is being
+used to render the styles. Sometimes, the order that the styles appear in the
+stylesheet matter for the semantics of the CSS. For instance, depending on the
+engine, the styles generated from
+
+```js
+const styles = StyleSheet.create({
+    ordered: {
+        margin: 0,
+        marginLeft: 15,
+    },
+});
+css(styles.ordered);
+```
+
+you might expect the following CSS to be generated:
+
+```css
+margin: 0px;
+margin-left: 15px;
+```
+
+but depending on the ordering of the keys in the style object, the CSS might
+appear as
+
+```css
+margin-left: 15px;
+margin: 0px;
+```
+
+which is semantically different, because the style which appears later will
+override the style before it.
+
+This might also manifest as a problem when server-side rendering, if the
+generated styles appear in a different order on the client and on the server.
+
+If you experience this issue where styles don't appear in the generated CSS in
+the order that they appear in your objects, there are two solutions:
+
+1. Don't use shorthand properties. For instance, in the margin example above,
+   by switching from using a shorthand property and a longhand property in the
+   same styles to using only longhand properties, the issue could be avoided.
+
+   ```js
+   const styles = StyleSheet.create({
+       ordered: {
+           marginTop: 0,
+           marginRight: 0,
+           marginBottom: 0,
+           marginLeft: 15,
+       },
+   });
+   ```
+
+2. Specify the ordering of your styles by specifying them using a
+   [`Map`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map).
+   Since `Map`s preserve their insertion order, Aphrodite is able to place your
+   styles in the correct order.
+
+   ```js
+   const styles = StyleSheet.create({
+       ordered: new Map([
+           ["margin", 0],
+           ["marginLeft", 15],
+       ]),
+   });
+   ```
+
+   Note that `Map`s are not fully supported in all browsers. It can be
+   polyfilled by using a package
+   like [es6-shim](https://www.npmjs.com/package/es6-shim).
+
 ## Advanced: Extensions
 
 Extra features can be added to Aphrodite using extensions.
--- a/package.json
+++ b/package.json
@@ -44,6 +44,7 @@
     "babel-core": "^5.8.25",
     "babel-loader": "^5.3.2",
     "chai": "^3.3.0",
+    "es6-shim": "^0.35.3",
     "eslint": "^3.7.1",
     "eslint-config-standard-react": "^4.2.0",
     "eslint-plugin-react": "^6.3.0",
--- a/src/generate.js
+++ b/src/generate.js
@@ -1,6 +1,7 @@
 /* @flow */
 import prefixAll from 'inline-style-prefixer/static';
 
+import OrderedElements from './ordered-elements';
 import {
     objectToPairs, kebabifyStyleName, recursiveMerge, stringifyValue,
     importantify, flatten
@@ -143,18 +144,21 @@ export const generateCSS = (
     stringHandlers /* : StringHandlers */,
     useImportant /* : boolean */
 ) /* : string */ => {
-    const merged = styleTypes.reduce(recursiveMerge);
+    const merged /* : OrderedElements */ = styleTypes.reduce(
+        recursiveMerge,
+        new OrderedElements());
 
-    const plainDeclarations = {};
+    const plainDeclarations = new OrderedElements();
     let generatedStyles = "";
 
-    Object.keys(merged).forEach(key => {
+    // TODO(emily): benchmark this to see if a plain for loop would be faster.
+    merged.forEach((key, val) => {
         // For each key, see if one of the selector handlers will handle these
         // styles.
         const foundHandler = selectorHandlers.some(handler => {
             const result = handler(key, selector, (newSelector) => {
                 return generateCSS(
-                    newSelector, [merged[key]], selectorHandlers,
+                    newSelector, [val], selectorHandlers,
                     stringHandlers, useImportant);
             });
             if (result != null) {
@@ -167,7 +171,7 @@ export const generateCSS = (
         // If none of the handlers handled it, add it to the list of plain
         // style declarations.
         if (!foundHandler) {
-            plainDeclarations[key] = merged[key];
+            plainDeclarations.set(key, val);
         }
     });
 
@@ -186,31 +190,25 @@ export const generateCSS = (
  * See generateCSSRuleset for usage and documentation of paramater types.
  */
 const runStringHandlers = (
-    declarations /* : SheetDefinition */,
+    declarations /* : OrderedElements */,
     stringHandlers /* : StringHandlers */,
     selectorHandlers /* : SelectorHandler[] */
 ) /* */ => {
-    const result = {};
-
     const hasStringHandlers = !!stringHandlers;
-    const keys = Object.keys(declarations);
-    for (let i = 0; i < keys.length; i += 1) {
+    return declarations.map((key, val) => {
         // If a handler exists for this particular key, let it interpret
         // that value first before continuing
-        if (hasStringHandlers && stringHandlers.hasOwnProperty(keys[i])) {
+        if (hasStringHandlers && stringHandlers.hasOwnProperty(key)) {
             // TODO(emily): Pass in a callback which generates CSS, similar to
             // how our selector handlers work, instead of passing in
             // `selectorHandlers` and have them make calls to `generateCSS`
             // themselves. Right now, this is impractical because our string
             // handlers are very specialized and do complex things.
-            result[keys[i]] = stringHandlers[keys[i]](
-                declarations[keys[i]], selectorHandlers);
+            return stringHandlers[key](val, selectorHandlers);
         } else {
-            result[keys[i]] = declarations[keys[i]];
+            return val;
         }
-    }
-
-    return result;
+    });
 };
 
 /**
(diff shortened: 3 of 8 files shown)


xymostech:
I haven't written docs/examples for this yet, just wanted to throw this out there to see if this sounds like a reasonable solution to the problem you're seeing in #199.

I don't think your code in particular would have to change in order for this to make your issue work correctly, since you're already implicitly specifying the ordering you want via two different style objects. If someone was having problems with e.g.

```
StyleSheet.create({
    a: {
        margin: 0,
        marginLeft: 15,
    },
})
```

turning into

```
margin-left: 15px;
margin: 0px;
```

then they could specify their styles via a `Map` like

```
StyleSheet.create({
    a: new Map([
        ["margin", 0],
        ["marginLeft", 15],
    ]),
})
```

Maybe that's not actually something we should worry about, but I don't know.

package.json
```diff
@@ -44,6 +44,7 @@
     "babel-core": "^5.8.25",
     "babel-loader": "^5.3.2",
     "chai": "^3.3.0",
+    "core-js": "^2.4.1",
```
lencioni: Are you bringing this in for Map? It might be a little bit nicer to bring in something a little more scoped like https://github.com/airbnb/browser-shims or just https://www.npmjs.com/package/es6-shim

src/generate.js
```diff
@@ -143,18 +144,20 @@ export const generateCSS = (
...
+    const merged /* : OrderedElements */ = styleTypes.reduce(
+        recursiveMerge,
+        new OrderedElements());
 
-    const plainDeclarations = {};
+    const plainDeclarations = new OrderedElements();
     let generatedStyles = "";
 
-    Object.keys(merged).forEach(key => {
+    merged.forEach((key, val) => {
```
lencioni: unrelated: Although I generally think stuff like `forEach` is best, regular for loops are faster which might matter in performance-sensitive applications. It might be worth looking into optimizing Aphrodite in this way at some point.

src/generate.js
```diff
@@ -244,44 +242,36 @@ const runStringHandlers = (
...
             return [[key, value]];
         })
     );
 
+    // Sort the prefixed rules according to the order that the keys were
+    // in originally before we prefixed them. Since the new, prefixed version
+    // of the rules aren't in the original list, we sort them to the top.
+    prefixedRules.sort((a, b) => {
+        const aIndex = handledDeclarations.keyOrder.indexOf(a);
+        const bIndex = handledDeclarations.keyOrder.indexOf(b);
```
lencioni: It would be nice to use something with constant time lookup here to avoid all of the array scans in this loop. A simple object would probably do the trick.

src/ordered-elements.js
```diff
@@ -0,0 +1,77 @@
...
+    get(key /* : string */) /* : any */ {
+        return this.elements[key];
+    }
+
+    has(key /* : string */) /* : boolean */ {
+        return this.elements.hasOwnProperty(key);
+    }
+}
+
+OrderedElements.fromObject = (obj) => {
```
lencioni: I wonder if this should take an optional array argument for `keyOrder` so it can be specified explicitly when possible.

lencioni:
Also, thanks for the fast response on this! 🎆

package.json
ljharb: `core-js`'s `Map` is not spec-compliant, since it mutates object keys in order to achieve otherwise-impossible performance. I highly recommend using the es6-shim instead here.

src/generate.js
ljharb: (please do so with benchmarks, however, and not just automatically)

package.json
xymostech: Interesting! This is specifically for the few tests which are using `Map`, not for the actual library, so I'm not super worried about spec compliance here. I can switch it out, though.

src/generate.js
xymostech: Yeah, that's definitely a good point. I'll put in a TODO to think about this later. We haven't done any serious benchmarking yet, so right now I think there are other places that could benefit more from optimization.

xymostech:
Thanks for the review, @lencioni! I'll look into making some of the changes you mentioned.

Not sure if this is possible for you, but would it be feasible for you to pull this change into your codebase to make sure that it actually fixes your problem? I'd hate to ship this only to see the bug still manifesting. :P

lencioni:
Yes, but I almost certainly won't get to it until Monday at the earliest

lencioni:
I've tested this out in my dev environment and it seems to resolve the issue we were running into!

src/ordered-elements.js
xymostech: At that point, you'd basically just be calling the constructor, so I'm not sure I see the usefulness of this. Anyway, we're not using that for now; we can add it if it's necessary.

src/generate.js
```diff
@@ -186,31 +190,25 @@ export const generateCSS = (
  * See generateCSSRuleset for usage and documentation of paramater types.
  */
 const runStringHandlers = (
-    declarations /* : SheetDefinition */,
+    declarations /* : OrderedElements */,
     stringHandlers /* : StringHandlers */,
     selectorHandlers /* : SelectorHandler[] */
 ) /* */ => {
-    const result = {};
```
xymostech: :( I undid some of your for-loop optimization here, but putting it back is lot uglier now with the `OrderedElements` thing. I re-did some of the optimizations inside of the `OrderedElements` implementation, though.

xymostech:
@lencioni @ljharb I think I addressed all of the things you mentioned, and I added some docs now too. Mind taking one more look at it before I land? I think once this and the rest of your diffs are landed @lencioni I'll release a new version.

README.md
```diff
@@ -429,6 +429,80 @@ then we end up with the opposite effect, with `foo` overriding `bar`! The way to
...
+   ```js
+   const styles = StyleSheet.create({
+       ordered: new Map([
+           ["margin", 0],
+           ["marginLeft", 15],
+       ]),
+   });
+   ```
+
+   Note that `Map`s are not fully supported in all browsers.
```
lencioni: It might be worth linking off to some resources here.

- https://github.com/airbnb/browser-shims
- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map

src/generate.js
lencioni: No worries, we can always come back, profile, and optimize the hotspots again. 🚀

src/generate.js
```diff
@@ -186,31 +190,25 @@ export const generateCSS = (
...
+        if (hasStringHandlers && stringHandlers.hasOwnProperty(key)) {
             // TODO(emily): Pass in a callback which generates CSS, similar to
             // how our selector handlers work, instead of passing in
             // `selectorHandlers` and have them make calls to `generateCSS`
             // themselves. Right now, this is impractical because our string
             // handlers are very specialized and do complex things.
-            result[keys[i]] = stringHandlers[keys[i]](
-                declarations[keys[i]], selectorHandlers);
+            return stringHandlers[key](val, selectorHandlers);
         } else {
```
lencioni: nit: you are returning early on 207, so you can remove the `else` here and move `return val;` up to the top level.

src/generate.js
```diff
@@ -246,44 +244,43 @@ const runStringHandlers = (
...
             return [[key, value]];
         })
     );
 
+    // Calculate the order that we want to each element in `prefixedRules` to
+    // be in, based on its index in the original key ordering.
+    const sortOrder = {};
+    for (let i = 0; i < prefixedRules.length; i++) {
+        const key = prefixedRules[i][0];
+        sortOrder[key] = handledDeclarations.keyOrder.indexOf(key);
```
lencioni: It would be nice to avoid the array scan in this loop too.

src/generate.js
```diff
@@ -246,44 +244,43 @@ const runStringHandlers = (
...
+    }
+
+    // Sort the prefixed rules according to the order that the keys were
+    // in originally before we prefixed them. Since the new, prefixed versions
+    // of the rules aren't in the original list, their sort order will be -1
+    // and we sort them to the top.
+    prefixedRules.sort((a, b) => {
+        return sortOrder[a[0]] - sortOrder[b[0]];
+    });
```
lencioni: I'm a little worried about this possibly changing the order of some prefixed rules in a problematic way. e.g. if you have a shorthand style and a longhand style of the same property that both need to be prefixed, will the prefixes preserve order? We might need to handle this a bit more delicately.

src/ordered-elements.js
```diff
@@ -0,0 +1,82 @@
...
+                callback(this.keyOrder[i], this.elements[this.keyOrder[i]])
+            );
+        }
+        return results;
+    }
+
+    set(key /* : string */, value /* : any */) {
+        if (!this.elements.hasOwnProperty(key)) {
+            this.keyOrder.push(key);
+        }
```
lencioni: If the element is already there, should it move it to the end in keyOrder?

src/ordered-elements.js
```diff
@@ -0,0 +1,82 @@
...
+    const ret = new OrderedElements();
+    map.forEach((val, key) => {
+        ret.set(key, val);
+    });
+    return ret;
+};
+
+OrderedElements.from = (obj) => {
+    if (obj instanceof OrderedElements) {
+        return obj;
```
lencioni: Should this create a new instance? That would be slower, but probably more correct.

README.md
xymostech: the mdn link is up above, I'll include a link to the shim here.

src/generate.js
xymostech: That's a good point. I'm not sure how to do this perfectly without changing the format that inline-style-prefixer outputs...

What if we do something like sort all of `key`, `"-webkit-" + key`, `"-moz-" + key`, and `"-ms-" + key` to the same position in the list? There are probably some prefixes we will miss which we'll put at the top, but that should catch most things?

src/ordered-elements.js
xymostech: I don't think so. Or at least, that would be semantically different than what we've had before. That is, merging

```
{
    a: 1,
    b: 2,
}
```

with

```
{
    a: 3,
}
```

gives

```
{
    a: 3,
    b: 2,
}
```

now. Moving to the end of the `keyOrder` would change the merged result to

```
{
    b: 2,
    a: 3,
}
```

lencioni: That's right. I'm wondering if the before behavior is unexpected and could cause similar bugs?

src/generate.js
lencioni: I think that might be okay as long as we only do this for prefixed styles that do not have a sortOrder. In other words, if I explicitly pass in a prefixed style after its unprefixed style, that order should be maintained.

xymostech: Good point! Thanks for catching that. Should be fixed in the newest version.

src/ordered-elements.js
xymostech: In the current way, it's impossible to make new styles appear after the old styles. Though if we switch it, it'll be impossible to make new styles appear before the old styles. I think either way has problems, but this at least stays consistent with what we have now.

PR #201, opened 2017-03-04
Improve `importantify()` performance

lencioni:
I was rolling through the code here looking for some opportunities to
speed it up and I noticed that there is a capture group in this regex
that isn't being used. By changing `(` to `(?:` we can avoid capturing
this group while still being able to use the parens for grouping.

I was also curious about the performance characteristics of various `replace`
arguments, and using a string in this way seems to have the edge, at least in
Firefox (35% faster) and Safari (12% faster)--Chrome performance seems
pretty similar for all possibilities. I also find this version more
readable, so it seems like we may as well go with this option.

https://jsperf.com/replace-string-vs-function/

I don't think we actually need a capture group at all here and can
accomplish the same thing by just anchoring on the end of the string.
This reduces the runtime of this function in my benchmark from ~105ms to
~35ms.

### DIFF
--- a/src/generate.js
+++ b/src/generate.js
@@ -284,12 +284,13 @@ export const generateCSSRuleset = (
         })
     );
 
-    const rules = prefixedRules.map(([key, value]) => {
-        const stringValue = (useImportant === false)
-            ? stringifyValue(key, value)
-            : importantify(stringifyValue(key, value));
-        return `${kebabifyStyleName(key)}:${stringValue};`;
-    }).join("");
+    const transformValue = (useImportant === false)
+        ? stringifyValue
+        : (key, value) => importantify(stringifyValue(key, value));
+
+    const rules = prefixedRules
+        .map(([key, value]) => `${kebabifyStyleName(key)}:${transformValue(key, value)};`)
+        .join("");
 
     if (rules) {
         return `${selector}{${rules}}`;
--- a/src/generate.js
+++ b/src/generate.js
@@ -285,9 +285,10 @@ export const generateCSSRuleset = (
     );
 
     const rules = prefixedRules.map(([key, value]) => {
-        const stringValue = stringifyValue(key, value);
-        const ret = `${kebabifyStyleName(key)}:${stringValue};`;
-        return useImportant === false ? ret : importantify(ret);
+        const stringValue = (useImportant === false)
+            ? stringifyValue(key, value)
+            : importantify(stringifyValue(key, value));
+        return `${kebabifyStyleName(key)}:${stringValue};`;
     }).join("");
 
     if (rules) {
--- a/src/util.js
+++ b/src/util.js
@@ -215,9 +215,10 @@ function murmurhash2_32_gc(str) {
 export const hashObject = (object /* : ObjectMap */) /* : string */ => murmurhash2_32_gc(JSON.stringify(object));
 
 
-const IMPORTANT_RE = /(?: !important)?$/;
-
 // Given a single style value string like the "b" from "a: b;", adds !important
 // to generate "b !important".
-export const importantify = (string /* : string */) /* : string */ =>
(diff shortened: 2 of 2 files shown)


xymostech:
This looks great! Thanks for looking in to this.

On my computer, the jsperf says that the hoisted function replacement is the fastest for both chrome and firefox:

image

image

What are you seeing?

lencioni:
screen shot 2017-03-04 at 3 46 06 pm
screen shot 2017-03-04 at 3 47 27 pm

But then I noticed my Firefox was pretty old so I updated. Here's the latest:

screen shot 2017-03-04 at 3 50 23 pm

So maybe function is the way to go?

lencioni:
On my phone (Pixel), function hoisted is the winner with function and string being slowest.

lencioni:
I think we can also simplify the regex here. If we change the code to this:

```js
const IMPORTANT_RE = /(?: !important)?;$/;

// Given a single style rule string like "a: b;", adds !important to generate
// "a: b !important;".
export const importantify = (string /* : string */) /* : string */ =>
    string.replace(IMPORTANT_RE, " !important;");
```

it drops the runtime of `importantify` from ~105ms to ~36ms.

lencioni:
FYI, I'm trying out all of my optimization PRs at the same time, and together they seem to speed up `css()` from 1.5s to 1.1s (~27%).

lencioni:
FYI, I also dug into inline-style-prefixer a little, and opened some PRs over there that should really help out as well.

- https://github.com/rofrischmann/inline-style-prefixer/pull/116
- https://github.com/rofrischmann/inline-style-prefixer/pull/115
- https://github.com/rofrischmann/css-in-js-utils/pull/1

src/util.js
```diff
@@ -211,11 +211,10 @@ function murmurhash2_32_gc(str) {
...
-export const importantify = (string /* : string */) /* : string */ =>
-    string.replace(
-        IMPORTANT_RE,
-        (_, base) => base + " !important;");
+// Given a single style value string like the "b" from "a: b;", adds !important
+// to generate "b !important".
+export const importantify = (string /* : string */) /* : string */ => (
+  string.slice(-11) === ' !important'
+    ? string
+    : `${string} !important`
```
xymostech: nit: we use 4 space indents

lencioni: Fixed!

src/util.js
```diff
@@ -216,11 +216,10 @@ function murmurhash2_32_gc(str) {
...
-// Given a single style rule string like "a: b;", adds !important to generate
-// "a: b !important;".
-export const importantify = (string /* : string */) /* : string */ =>
-    string.replace(
-        IMPORTANT_RE,
-        (_, base) => base + " !important;");
+// Given a single style value string like the "b" from "a: b;", adds !important
+// to generate "b !important".
+export const importantify = (string /* : string */) /* : string */ => (
+    string[string.length - 10] === '!' && string.slice(-11) === ' !important'
```
xymostech: hehe this is great! Could you maybe add a comment explaining why this check looks so funky? Also maybe wrap these checks in parens because I can never remember the precedence of `?:`? :P

src/generate.js
```diff
@@ -286,8 +286,7 @@ export const generateCSSRuleset = (
 
     const rules = prefixedRules.map(([key, value]) => {
         const stringValue = stringifyValue(key, value);
-        const ret = `${kebabifyStyleName(key)}:${stringValue};`;
-        return useImportant === false ? ret : importantify(ret);
+        return `${kebabifyStyleName(key)}:${useImportant === false ? stringValue : importantify(stringValue)};`;
```
xymostech: This line is a bit long. :( Maybe split it up into two?

lencioni: Done!

src/util.js
lencioni: Done!

PR #203, opened 2017-03-04
Replace murmur hash with djb2 hash

lencioni:
In profiling StyleSheet.create, I noticed that much of the time was
spent hashing. So, I found a faster hashing algorithm.

The implementation was taken from:

  https://github.com/darkskyapp/string-hash

According to this StackExchange post, this algorithm doesn't have as
good of randomness, but it has about the same percentage of collisions.
I don't think randomness matters for this application, so I think this
is okay.

http://softwareengineering.stackexchange.com/a/145633

Using similar methodology to #202, this appears to make StylSheet.create
~15% faster (~220ms to ~185ms).

cc @ljharb

### DIFF
--- a/src/util.js
+++ b/src/util.js
@@ -156,52 +156,23 @@ export const stringifyValue = (
 };
 
 /**
- * JS Implementation of MurmurHash2
+ * JS Implementation of djb2
  *
- * @author <a href="mailto:gary.court@gmail.com">Gary Court</a>
- * @see http://github.com/garycourt/murmurhash-js
- * @author <a href="mailto:aappleby@gmail.com">Austin Appleby</a>
- * @see http://sites.google.com/site/murmurhash/
- *
- * @param {string} str ASCII only
- * @return {string} Base 36 encoded hash result
+ * @see https://github.com/darkskyapp/string-hash
  */
-function murmurhash2_32_gc(str) {
-    let l = str.length;
-    let h = l;
-    let i = 0;
-    let k;
 
-    while (l >= 4) {
-        k = ((str.charCodeAt(i) & 0xff)) |
-            ((str.charCodeAt(++i) & 0xff) << 8) |
-            ((str.charCodeAt(++i) & 0xff) << 16) |
-            ((str.charCodeAt(++i) & 0xff) << 24);
+function djb2Hash(str) {
+  let hash = 5381;
+  let i = str.length;
 
-        k = (((k & 0xffff) * 0x5bd1e995) + ((((k >>> 16) * 0x5bd1e995) & 0xffff) << 16));
-        k ^= k >>> 24;
-        k = (((k & 0xffff) * 0x5bd1e995) + ((((k >>> 16) * 0x5bd1e995) & 0xffff) << 16));
+  while(i) {
+    hash = (hash * 33) ^ str.charCodeAt(--i);
+  }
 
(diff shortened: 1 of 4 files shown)


lencioni:
Also, I didn't look very hard for a faster hashing algorithm, so if you know of one that's even faster, let's try that!

src/util.js
```diff
@@ -152,52 +152,23 @@ export const stringifyValue = (
...
-    /* eslint-disable no-fallthrough */ // forgive existing code
-    switch (l) {
-    case 3: h ^= (str.charCodeAt(i + 2) & 0xff) << 16;
-    case 2: h ^= (str.charCodeAt(i + 1) & 0xff) << 8;
-    case 1: h ^= (str.charCodeAt(i) & 0xff);
-        h = (((h & 0xffff) * 0x5bd1e995) + ((((h >>> 16) * 0x5bd1e995) & 0xffff) << 16));
-    }
-    /* eslint-enable no-fallthrough */
+function djb2Hash(str) {
+  let hash = 5381;
```
ljharb: what's this magic number?

lencioni: It was apparently chosen through manual testing as the number that produced fewer collisions and better avalanching.

http://stackoverflow.com/a/10697529/18986
https://groups.google.com/forum/#!msg/comp.lang.c/lSKWXiuNOAk/zstZ3SRhCjgJ
http://stackoverflow.com/a/4825477/18986

lencioni:
JSPerf: https://jsperf.com/murmurhash2-vs-djb2

This implementation of djb2 seems to be about 2x faster than the implementation of MurmurHash2 in Chrome. In Firefox, they seem to be pretty equivalent.

csilvers:
I should comment on this I guess, since I'm the one who pushed for murmurhash.  It's a very fast (and good) hash when implemented in C, but maybe not in javascript.  I don't have a problem moving to djb2, though it's not a great hash function.  (The claim about "same number of collisions" depends on the dataset.  I doubt it's true in general, though may be true for the types of strings we're hashing in aphrodite -- that could be tested experimentally, though may not be worth the trouble.  After all, if a hash is twice as fast that can pay for a lot of collision-handling code!)

The best thing to do would be to ask the js engine to use their internal hash function, something like `hash` in python.  (It looks like chrome uses jenkinshash -- which is a predecessor to murmurhash and is pretty good -- http://stackoverflow.com/questions/30695564/what-kind-of-hashing-function-algorithm-does-javascript-associative-array-use.)  I don't know if javascript exposes that though.

lencioni:
> The best thing to do would be to ask the js engine to use their internal hash function

This would be great, but I don't think it is an option. And, if we could do this, it would have to be the same hash function across all browsers to ensure that the hash is the same from server -> client.

ljharb:
JS definitely doesn't expose anything like that.

src/util.js
```diff
@@ -212,7 +164,7 @@ function murmurhash2_32_gc(str) {
 // this to produce consistent hashes browsers need to have a consistent
 // ordering of objects. Ben Alpert says that Facebook depends on this, so we
 // can probably depend on this too.
-export const hashObject = (object /* : ObjectMap */) /* : string */ => murmurhash2_32_gc(JSON.stringify(object));
+export const hashObject = (object /* : ObjectMap */) /* : number */ => stringHash(JSON.stringify(object));
```
xymostech: Could you stringify the output here in the same way that we did in murmurhash? I think a `.toString(36)` at the end here is all that's needed.

lencioni: Done! I also profiled this and it seems to be equivalent to not converting to a string.

PR #205, opened 2017-03-05
Update inline-style-prefixer ^2.0.0 -> ^3.0.1

lencioni:
It seems that this version is a complete rewrite with some performance
improvements. In my benchmarks, this seems to improve the speed of
`css()` by about 10%.

Changelog:

* performance improvements (~10% faster)
* ordering prefixed properties correctly
* introducing new core prefixer that perform up to 4.5x faster
* added a whole new generator to create your custom prefixer
* added 4 new plugins to prefix special values
* new documentation using gitbook
* integrated flowtype


lencioni:
I've been seeing these test failures locally even before this update. I wonder if a change in an earlier version of inline-style-prefixer that wasn't being run in CI caused this.

lencioni:
I tried v2.0.0 locally and I still get this failure. Perhaps it is spurious in CI? Maybe it is related to the node version or a deep dependency difference?

lencioni:
I'm trying this out again on a different laptop and I am able to reproduce this locally before and after this change. This looks like it might be a bug in inline-style-prefixer. It looks like inside of `generateCSSRuleset`, `prefixAll` when given `{ transition: 'border-color 200ms linear' }` returns 

```js
{ transition: 'border-color 200ms linear',
  WebkitTransition: 'border-color 200ms linear',
  MozTransition: 'border-color 200ms linear' }
```

which is not the order that we want.

Also, it seems that `prefixAll` mutates its argument, which I think might be okay for us here but was a bit unexpected to me.

@rofrischmann can you take a look at this?

rofrischmann:
Yeah, we do not explicitly order prefixes as we would have to populate the whole object again. We might provide a separate function to do that, but I don't want it to be part of the core as there actually is nothing wrong with the above mentioned order.

lencioni:
@rofrischmann the order of these is important, since the vendor prefixed versions can do different things. See https://css-tricks.com/ordering-css3-properties/

lencioni:
@rofrischmann I think that @xymostech is hoping to get a new release of Aphrodite out today and it would be really great to include this update as well. Any chance you can publish a new version?

rofrischmann:
@lencioni @xymostech I've been away yesterday. I can do that today, but later today. Maybe 5pm, its 10.34am right now. (Germany)

lencioni:
Thanks @rofrischmann! :man_dancing:

lencioni:
Sadly, 3.0.1 didn't actually fix the issue we are running into here. It seems that `style` has already been prefixed by the time it hits `prefixProperty`. I'll keep looking into this.

xymostech:
Thanks for looking into this @lencioni!! It looks like this isn't going to get out soon enough for the upcoming release, but we can keep track of it for the future. :)

rofrischmann:
I think we're not changing that. Immutable objects are just so much slower here. You can easily add another function that sorts the keys once. But I do not want it to be default behavior.

Really, I never had any issues with "incorrect" order. Thinks like border-radius e.g. don't get prefixed anyways.

lencioni:
I wonder if this will perhaps not be an issue after #200 lands. @xymostech I'm happy to try rebasing this once you merge your PR.

lencioni:
@xymostech I've updated this PR on top of your change, and have modified the tests so they pass. It seems that the new version of inline-style-prefixer stops prefixing some things, so PTAL and pay attention to the changes I've made to the tests.

tests/generate_test.js
```diff
@@ -161,16 +190,15 @@ describe('generateCSS', () => {
             WebkitAlignItems: 'center',
             justifyContent: 'center',
         }], '.foo{' +
-            '-ms-flex-pack:center !important;' +
-            '-webkit-box-align:center !important;' +
-            '-ms-flex-align:center !important;' +
             '-webkit-box-pack:center !important;' +
-            'display:flex !important;' +
```
xymostech: Oh actually now that I look at this more closely, it seems like `display: flex` should go after `display: -webkit-box`? Maybe we should ensure that. I can do that separately from this change, though.

Other than that, all of these changes seem quite reasonable to me.

tests/generate_test.js
```diff
@@ -199,6 +227,10 @@ describe('generateCSS', () => {
     it('correctly prefixes border-color transition properties', () => {
       assertCSS('.foo', [{
         'transition': 'border-color 200ms linear'
-      }], '.foo{-webkit-transition:border-color 200ms linear !important;transition:border-color 200ms linear !important;}');
+      }], '.foo{' +
+          '-webkit-transition:border-color 200ms linear !important;' +
+          '-moz-transition:border-color 200ms linear !important;' +
+          'transition:border-color 200ms linear !important;' +
+      '}');
```
xymostech: <3 formatting

rofrischmann:
You can generate your own prefixer version if you need older browser support ;)

tests/generate_test.js
```diff
@@ -10,7 +10,21 @@ describe('generateCSSRuleset', () => {
...
+            `
+Expected:
+
+${formatStyles(expectedNormalized)}
+
+Actual:
+
+${formatStyles(actual)}
+`
+        );
```
xymostech: Can I see an example of what an error looks like with this message?

xymostech:
Looks like maybe there's a guide here? https://github.com/rofrischmann/inline-style-prefixer/blob/master/docs/guides/CustomPrefixAll.md

tests/generate_test.js
lencioni:
```
  2) generateCSS correctly prefixes border-color transition properties:
     AssertionError:
Expected:

.foo{
-webkit-transition:border-color 200ms linear !important;
transition:border-color 200ms linear !important;
}


Actual:

.foo{
-webkit-transition:border-color 200ms linear !important;
-moz-transition:border-color 200ms linear !important;
transition:border-color 200ms linear !important;
}

: expected '.foo{-webkit-transition:border-color 200ms linear !important;-moz-transition:border-color
 200ms linear !important;transition:border-color 200ms linear !important;}' to equal '.foo{-webkit-tr
ansition:border-color 200ms linear !important;transition:border-color 200ms linear !important;}'
      at assertCSS (tests/generate_test.js:98:8)
      at Context.<anonymous> (tests/generate_test.js:204:4)
```

tests/generate_test.js
lencioni: Yep, we need `display: flex` to be after the webkit one.

lencioni:
@xymostech agreed, it would be best to keep the same browser support. If you update this object to match the versions supported by the current Aphrodite, I'll take a swing at generating our own prefixAll.
```js
const browserList = {
  chrome: 46,
  android: 4,
  firefox: 40,
  ios_saf: 8,
  safari: 8,
  ie: 11,
  ie_mob: 11,
  edge: 12,
  opera: 16,
  op_mini: 12,
  and_uc: 9,
  and_chr: 46
}
```

xymostech:
Here's a patch that adds the custom prefixAll: https://gist.github.com/xymostech/e01d41c09397b8e4cf4c6b116cc9c9be

I mostly just changed `ie` to `9` though, I haven't really thought through which versions we want to support...

lencioni:
Oh sweet. It seems like you've got it under control then. Let me know if you'd like me to make any changes to this PR or if you want to merge and follow up with the browser version fixes.

xymostech:
I think if you just add that patch to this commit that would be good?

Now that I think about it, maybe we could just set all of the versions to `0` or `1` or something, since we just sorta want all of the prefixes that are available. Testing that out locally, it seems to do what we want (all the current prefixes plus the old IE ones). Does that sound good? (I updated the patch appropriately)

lencioni:
I think that sounds like an okay place to start for now. It would be nice to eventually allow folks to customize this. I'll incorporate your patch shortly.

lencioni:
@xymostech I just pushed a new commit that includes your patch. I had to make a couple of changes to get it to work. PTAL!

tests/generate_test.js
xymostech: Awesome.

xymostech:
Looks fine, but I guess now the travis tests are failing. Oops. Maybe add `tools/generate_prefixer_data.js` to the `pretest` script?

lencioni:
@xymostech done!

xymostech:
Awesome. I'm gonna take a quick look into moving `display: flex` below `display: -webkit-flex` and then release version 1.2.0. Sound good?

lencioni:
That sounds awesome. I'm really excited to get this into production and see if it has any impact on our performance metrics.

rofrischmann:
https://github.com/rofrischmann/inline-style-prefixer/blob/5f456345960512063cd4e3a0f6984fa1ce8951fb/config.js These are the versions, the previous majors 1.x.x and 2.x.x supported. I think you're safe using those. Otherwise you will get a lot of useless prefixes such as `borderRadius` which really bloats your CSS output.

xymostech:
Oh, thank you @rofrischmann! I wasn't sure where that file lived before. I'll switch to using those versions, since that's actually what we supported before.

PR #207, opened 2017-03-05
Optimize kebabify

lencioni:
By passing a function to replace, we can do the uppercase to lowercase
conversion at the same time. This is pretty much the exact example that
MDN gives, with the only difference being that we aren't checking
against the offset here to avoid adding "-" at the beginning of the
string:

  https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replace#Using_an_inline_function_that_modifies_the_matched_characters

In my profiling, this seems to make kebabify 50% faster (50ms -> 25ms).

### DIFF
--- a/src/util.js
+++ b/src/util.js
@@ -33,13 +33,14 @@ export const flattenDeep = (list /* : any[] */) /* : any[] */ =>
 
 const UPPERCASE_RE = /([A-Z])/g;
 const UPPERCASE_RE_TO_KEBAB = (match /* : string */)  /* : string */ => `-${match.toLowerCase()}`;
-const MS_RE = /^ms-/;
 
-export const kebabifyStyleName = (string /* : string */) /* : string */ => (
-  string
-    .replace(UPPERCASE_RE, UPPERCASE_RE_TO_KEBAB)
-    .replace(MS_RE, '-ms-')
-);
+export const kebabifyStyleName = (string /* : string */) /* : string */ => {
+    const result = string.replace(UPPERCASE_RE, UPPERCASE_RE_TO_KEBAB);
+    if (result[0] === 'm' && result[1] === 's' && result[2] === '-') {
+        return `-${result}`;
+    }
+    return result;
+};
 
 const isNotObject = (
   x/* : ObjectMap | any */
--- a/src/util.js
+++ b/src/util.js
@@ -35,8 +35,11 @@ const UPPERCASE_RE = /([A-Z])/g;
 const UPPERCASE_RE_TO_KEBAB = (match /* : string */)  /* : string */ => `-${match.toLowerCase()}`;
 const MS_RE = /^ms-/;
 
-const kebabify = (string /* : string */) /* : string */ => string.replace(UPPERCASE_RE, UPPERCASE_RE_TO_KEBAB);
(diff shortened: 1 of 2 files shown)


src/util.js
```diff
@@ -32,10 +32,15 @@ export const flattenDeep = (list /* : any[] */) /* : any[] */ =>
...
-const kebabify = (string /* : string */) /* : string */ => string.replace(UPPERCASE_RE, '-$1').toLowerCase();
-export const kebabifyStyleName = (string /* : string */) /* : string */ => kebabify(string).replace(MS_RE, '-ms-');
+const UPPERCASE_RE_TO_KEBAB = (match /* : string */)  /* : string */ => `-${match.toLowerCase()}`;
+
+export const kebabifyStyleName = (string /* : string */) /* : string */ => {
+  const result = string.replace(UPPERCASE_RE, UPPERCASE_RE_TO_KEBAB);
+  if (result.slice(0, 3) === 'ms-') {
+    return `-${result}`;
+  }
+  return result;
```
xymostech: nit: we use 4-space indents

src/util.js
```diff
@@ -32,10 +32,15 @@ export const flattenDeep = (list /* : any[] */) /* : any[] */ =>
...
 const UPPERCASE_RE = /([A-Z])/g;
-const MS_RE = /^ms-/;
-
-const kebabify = (string /* : string */) /* : string */ => string.replace(UPPERCASE_RE, '-$1').toLowerCase();
-export const kebabifyStyleName = (string /* : string */) /* : string */ => kebabify(string).replace(MS_RE, '-ms-');
+const UPPERCASE_RE_TO_KEBAB = (match /* : string */)  /* : string */ => `-${match.toLowerCase()}`;
+
+export const kebabifyStyleName = (string /* : string */) /* : string */ => {
+  const result = string.replace(UPPERCASE_RE, UPPERCASE_RE_TO_KEBAB);
+  if (result.slice(0, 3) === 'ms-') {
```
xymostech: Is this the fastest way to check this? It seems like checking `result[0] === "ms" && result[1] === "s" && result[2] === "-"` might be faster than slicing the string, but is also really ugly. Maybe that's what [`.startsWith()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/startsWith) does, but not everything supports it yet...

lencioni: This microbenchmark shows bracket is probably the way to go. Not sure if it will really matter much in practice, but it is simple enough so I'll make the change.
https://jsperf.com/slice-vs-bracket-string-check/1

src/util.js
lencioni: Done!

src/util.js
xymostech: I made a maybe-more-useful microbenchmark that actually tests strings that start with `ms-` and it also agrees (much more strongly actually) so this looks great! :) https://jsperf.com/slice-vs-bracket-string-check-ms

lencioni: Ah, in that case, maybe we should use a similar approach in importantify?

lencioni: I added a similar optimization to my importantify PR: https://github.com/Khan/aphrodite/pull/201/commits/bd0a575916b7210847147a79f775c24199b5572a

PR #208, opened 2017-03-05
Replace forEach with for loop in runStringHandlers

lencioni:
According to my profiling, this seems to make runStringHandlers about
10% faster.

### DIFF
--- a/src/generate.js
+++ b/src/generate.js
@@ -192,21 +192,22 @@ const runStringHandlers = (
 ) /* */ => {
     const result = {};
 
-    Object.keys(declarations).forEach(key => {
+    const keys = Object.keys(declarations);
+    for (let i = 0; i < keys.length; i += 1) {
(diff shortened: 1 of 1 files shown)


src/generate.js
```diff
@@ -192,21 +192,23 @@ const runStringHandlers = (
 ) /* */ => {
     const result = {};
 
-    Object.keys(declarations).forEach(key => {
+    const hasStringHandlers = !!stringHandlers;
+    const keys = Object.keys(declarations);
+    for (var i = 0; i < keys.length; i += 1) {
```
xymostech: `let i = 0` maybe? Is that slower?

lencioni: Oops! Should be the same speed, since I think babel compiles it down to `var`. I think all of the error checking related to the block scoping happens at build time.

lencioni: Fixed!

PR #211, opened 2017-03-08
Sort prefixed values before unprefixed values with the same key

xymostech:
Summary: In #200 the way we started sorting prefixed and unprefixed values
differently. This builds on that by making sure that when style values are
prefixed, they come before the unprefixed values with the same key.

E.g.

```css
display: -webkit-flex; // prefixed value comes before
display: flex;         // unprefixed value
```

@lencioni

### DIFF
--- a/src/generate.js
+++ b/src/generate.js
@@ -255,31 +255,7 @@ export const generateCSSRuleset = (
     const handledDeclarations /* : OrderedElements */ = runStringHandlers(
         declarations, stringHandlers, selectorHandlers);
 
-    // Calculate the order that we want to each element in `prefixedRules` to
-    // be in, based on its index in the original key ordering. We have to do
-    // this before the prefixing, because prefixAll mutates
-    // handledDeclarations.elements.
-    const sortOrder = {};
-    for (let i = 0; i < handledDeclarations.keyOrder.length; i++) {
-        const key = handledDeclarations.keyOrder[i];
-        sortOrder[key] = i;
-
-        // In order to keep most prefixed versions of keys in about the same
-        // order that the original keys were in but placed before the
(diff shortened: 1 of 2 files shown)


src/generate.js
```diff
@@ -255,10 +255,25 @@ export const generateCSSRuleset = (
     const handledDeclarations /* : OrderedElements */ = runStringHandlers(
         declarations, stringHandlers, selectorHandlers);
 
+    const originalElements = {...handledDeclarations.elements};
+
+    // NOTE(emily): This mutates handledDeclarations.elements.
+    const prefixedDeclarations = prefixAll(handledDeclarations.elements);
+
+    const prefixedRules = flatten(
+        objectToPairs(prefixedDeclarations).map(([key, value]) => {
```
lencioni: unrelated: I wonder if we should replace objectToPairs with Object.entries

src/generate.js
```diff
@@ -275,35 +290,38 @@ export const generateCSSRuleset = (
...
-    // Sort the prefixed rules according to the order that the keys were
-    // in originally before we prefixed them. New, prefixed versions
-    // of the rules aren't in the original list, so we set their order to -1 so
-    // they sort to the top.
+    // Actually sort the rules according to the sort order.
     prefixedRules.sort((a, b) => {
-        const aOrder = sortOrder.hasOwnProperty(a[0]) ? sortOrder[a[0]] : -1;
-        const bOrder = sortOrder.hasOwnProperty(b[0]) ? sortOrder[b[0]] : -1;
-        return aOrder - bOrder;
+        return sortOrderForProperty(a) - sortOrderForProperty(b);
```
lencioni: nit: this could be simplified to use the implicit return style

PR #213, opened 2017-03-08
Optimize injectAndGetClassName

lencioni:
This function is called every time css() is used, so we want it to be as
fast as possible. I spotted three loops here that could be condensed
into one. In my profiling this seems to make the function about 30%
faster.

### DIFF
--- a/src/inject.js
+++ b/src/inject.js
@@ -219,19 +219,29 @@ export const injectAndGetClassName = (
 ) /* : string */ => {
     styleDefinitions = flattenDeep(styleDefinitions);
 
-    // Filter out falsy values from the input, to allow for
-    // `css(a, test && c)`
-    const validDefinitions = styleDefinitions.filter((def) => def);
-
+    const classNameBits = [];
+    const definitionBits = [];
+    for (let i = 0; i < styleDefinitions.length; i += 1) {
+        // Filter out falsy values from the input, to allow for
+        // `css(a, test && c)`
+        if (styleDefinitions[i]) {
+            classNameBits.push(styleDefinitions[i]._name);
+            definitionBits.push(styleDefinitions[i]._definition);
+        }
+    }
     // Break if there aren't any valid styles.
-    if (validDefinitions.length === 0) {
+    if (classNameBits.length === 0) {
         return "";
     }
+    const className = classNameBits.join("-o_O-");
 
-    const className = validDefinitions.map(s => s._name).join("-o_O-");
-    injectStyleOnce(className, `.${className}`,
-        validDefinitions.map(d => d._definition),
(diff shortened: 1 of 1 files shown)


src/inject.js
```diff
@@ -219,19 +219,29 @@ export const injectAndGetClassName = (
...
-    if (validDefinitions.length === 0) {
+    if (classNameBits.length === 0) {
         return "";
     }
-
-    const className = validDefinitions.map(s => s._name).join("-o_O-");
-    injectStyleOnce(className, `.${className}`,
-        validDefinitions.map(d => d._definition),
-        useImportant, selectorHandlers);
+    const className = `${classNameBits.join("-o_O-")}`
```
xymostech: This doesn't need to be in a template string, right?

src/inject.js
```diff
@@ -219,19 +219,29 @@ export const injectAndGetClassName = (
 ) /* : string */ => {
     styleDefinitions = flattenDeep(styleDefinitions);
 
-    // Filter out falsy values from the input, to allow for
-    // `css(a, test && c)`
-    const validDefinitions = styleDefinitions.filter((def) => def);
-
+    let classNameBits = [];
+    let definitionBits = [];
```
xymostech: These can be `const`, right?

src/inject.js
```diff
@@ -219,19 +219,29 @@ export const injectAndGetClassName = (
...
+    for (let i = 0; i < styleDefinitions.length; i += 1) {
+      // Filter out falsy values from the input, to allow for
+      // `css(a, test && c)`
+      if (styleDefinitions[i]) {
+        classNameBits.push(styleDefinitions[i]._name);
+        definitionBits.push(styleDefinitions[i]._definition);
+      }
+    }
```
xymostech: nit: 4 space indents, please

src/inject.js
lencioni: Oh yes, of course.

src/inject.js
lencioni: gah! Would you be opposed to me setting up eslint to tell me about these things for this project?

src/inject.js
lencioni: oops, nope.

lencioni:
Updated!

src/inject.js
xymostech: Hahaha not at all, I was thinking the same thing. We have eslint but it doesn't have much turned on. :P I think ideally it would follow the KA style eslint at https://github.com/Khan/khan-linter/blob/master/eslintrc but that's ginormous so maybe just adding in a few of the common annoying ones would be good...

xymostech:
Mmkay, gonna go make the release now. Wish me luck!

PR #214, opened 2017-03-08
Optimize flattenDeep

lencioni:
This is called by injectAndGetClassName which is called by css, so we
want it to be as fast as possible. The array-flatten package serves the
same purpose and has been well-optimized. Dropping that in here, makes
this go much faster, about 2-3 times faster in my profiling.

@xymostech

### DIFF
--- a/package.json
+++ b/package.json
@@ -61,6 +61,7 @@
     "webpack": "^1.12.2"
   },
   "dependencies": {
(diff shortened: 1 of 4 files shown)


lencioni:
Closing this because I have something better in 2325e6d36308e75e6191cff26001d9aa441a355e, #216.

PR #216, opened 2017-03-10
Optimize css()

lencioni:
After the 1.2.0 release, I noticed that most of my optimizations were
completely offset by the ordering bugfixes that were added. After some
profiling, I noticed that generateCSSRuleset was ripe for optimization.

My approach here is to reduce the amount of work that needs to be done
to a bare minimum. I accomplish this by being smarter about when to loop
and how many loops to run. In my benchmark, this PR reduces the runtime
of css() from 2750 ms to 900 ms, which is 3.05x faster.

Some of this performance improvement will only be seen when running on the client,
but on the server this should still result in ~2x faster improvement.

This is probably best reviewed commit-by-commit.

Before:

screen shot 2017-03-21 at 2 31 50 pm

After:

screen shot 2017-03-21 at 2 31 26 pm

@xymostech


src/generate.js
```diff
@@ -252,87 +265,79 @@ export const generateCSSRuleset = (
...
+                    // This is a Moz-prefixed style, like "MozTransition". We check
+                    // the second character to avoid colliding with Ms-prefixed
+                    // styles. Let's find its original style's sort order.
+                    originalStyle = elementNames[i][3].toLowerCase() + elementNames[i].slice(4);
+                } else if (elementNames[i][1] === 's') {
+                    // This is a Ms-prefixed style, like "MsTransition". We check
+                    // the second character to avoid colliding with Moz-prefixed
+                    // styles. Let's find its original style's sort order.
+                    originalStyle = elementNames[i][2].toLowerCase() + elementNames[i].slice(3);
+                }
```
lencioni: Looks like coverage is failing branch coverage for the `else` case here. Any thoughts on how to fix this?

lencioni: I decided to change it from an else if to an else.

src/ordered-elements.js
```diff
@@ -1,5 +1,4 @@
 /* @flow */
-/* global Map */
 
 export default class OrderedElements {
```
lencioni: The branch coverage on this file falls to 75%, but I can't see any missed branches in the HTML report. I'm guessing it might not be getting the coverage right with the Babel transform happening here. Thoughts?

lencioni: Figured it out. The constructor was never called with arguments, so I removed the args.

iancmyers:
Super excited for this!

lencioni:
@xymostech Are there any changes you'd like me to make to this PR before merging?

xymostech:
@lencioni Sorry, I'd like to review this before merging but I've just been terribly busy. I'll review this today or over the weekend.

xymostech:
Hi! My computer broke over the weekend, so I didn't get to reviewing this... It's fixed now, I'll review this today.

lencioni:
@xymostech no worries! I'm sorry to hear about your computer troubles. :( Glad to hear that you are back in action.

src/generate.js
```diff
@@ -198,12 +198,15 @@ const runStringHandlers = (
     declarations /* : OrderedElements */,
     stringHandlers /* : StringHandlers */,
     selectorHandlers /* : SelectorHandler[] */
-) /* */ => {
-    const hasStringHandlers = !!stringHandlers;
+) /* : OrderedElements */ => {
```
xymostech: <3 for the flow fix

src/generate.js
```diff
@@ -203,20 +203,18 @@ const runStringHandlers = (
         return declarations;
     }
 
-    return declarations.map((key, val) => {
-        // If a handler exists for this particular key, let it interpret
```
xymostech: could you add these comments back into the code?

.eslintrc.json
```diff
@@ -5,7 +5,9 @@
   ],
   "env": {
     "browser": true,
+    "es6": true,
     "node": true
+
```
xymostech: nit: remove this empty line

package.json
```diff
@@ -49,7 +49,6 @@
     "caniuse-api": "^1.5.3",
     "chai": "^3.3.0",
     "coveralls": "^2.12.0",
-    "es6-shim": "^0.35.3",
```
xymostech: why did this go away? Are we not still testing if `Map` styles work correctly?

src/inject.js
```diff
@@ -154,13 +160,15 @@ export const injectStyleOnce = (
     useImportant /* : boolean */,
     selectorHandlers /* : SelectorHandler[] */ = []
 ) => {
-    if (!alreadyInjected[key]) {
-        const generated = generateCSS(
-            selector, definitions, selectorHandlers,
-            stringHandlers, useImportant);
-
-        injectGeneratedCSSOnce(key, generated);
+    if (alreadyInjected[key]) {
```
xymostech: Does changing the structure of this function actually make it faster?

src/inject.js
```diff
@@ -217,28 +244,23 @@ export const injectAndGetClassName = (
     styleDefinitions /* : MaybeSheetDefinition[] */,
     selectorHandlers /* : SelectorHandler[] */
 ) /* : string */ => {
-    styleDefinitions = flattenDeep(styleDefinitions);
+    const processedStyleDefinitions /* ProcessedStyleDefinitions */ = {
```
xymostech: nit: I think you need a `:` before `ProcessedStyleDefinitions` here in order to make flow actually notice this.

src/util.js
```diff
@@ -21,9 +21,6 @@ export const mapObj = (
     return mappedObj;
 }
 
-export const flattenDeep = (list /* : any[] */) /* : any[] */ =>
-    list.reduce((memo, x) => memo.concat(Array.isArray(x) ? flattenDeep(x) : x), []);
```
xymostech: oh yeah, I forgot how horribly inefficient this function was. Thanks!

package.json
```diff
@@ -56,13 +56,15 @@
...
     "mocha": "^2.3.3",
+    "mock-require": "^2.0.1",
     "npm-run-all": "^1.7.0",
     "nyc": "^6.4.4",
     "rimraf": "^2.5.2",
     "webpack": "^1.12.2"
   },
   "dependencies": {
     "asap": "^2.0.3",
+    "can-use-dom": "^0.1.0",
```
xymostech: how big is this dependency? Can we just replace this with a `typeof document !== "undefined"` check like we do in other places? (or replace those other checks with this?)

src/prefix.js
```diff
@@ -0,0 +1,27 @@
...
+
+if (canUseDOM) {
+    // Client-side, we can use the dynamic prefixer to avoid unnecessary work.
+    const Prefixer = require('inline-style-prefixer');
+
+    // We need keepUnprefixed so our sorting code knows how to order the
+    // prefixed styles.
+    const prefixer = new Prefixer({
+        keepUnprefixed: true,
+        //userAgent: navigator.userAgent,
```
xymostech: nit: this is commented out.

src/prefix.js
```diff
@@ -0,0 +1,27 @@
...
+
+let prefix;
+
+if (canUseDOM) {
+    // Client-side, we can use the dynamic prefixer to avoid unnecessary work.
+    const Prefixer = require('inline-style-prefixer');
+
+    // We need keepUnprefixed so our sorting code knows how to order the
+    // prefixed styles.
+    const prefixer = new Prefixer({
```
xymostech: Aside from server-side rendering, I know one of the reasons we initially didn't go with this approach is because the code to determine which prefixes were needed (which we're doing a build time in `staticPrefixData`) took up a lot of space. Can you build aphrodite before and after this change and see how much bigger the resulting dist file is?

tests/prefix_test.js
```diff
@@ -0,0 +1,63 @@
...
+        });
+    });
+
+    /* istanbul ignore next */
+    describe('on the client', () => {
+        const originalNavigator = global.navigator;
+        const originalUserAgent = global.navigator && global.navigator.userAgent;
+        let prefix;
+
+        beforeEach(() => {
```
xymostech: <3 great tests

src/generate.js
```diff
@@ -258,81 +270,71 @@ export const generateCSSRuleset = (
...
+        // Loop through prefixedElements, looking for anything that is not in
+        // sortOrder, which means it was added by prefixAll. This means that we
+        // need to figure out where it should appear in the sortOrder.
+        for (let i = 0; i < elementNames.length; i++) {
+            if (!originalElements.hasOwnProperty(elementNames[i])) {
+                // This element is not in the sortOrder, which means it is a prefixed
+                // value that was added by prefixAll. Let's try to figure out where it
+                // goes.
+                let originalStyle;
+                if (elementNames[i][0] === 'W') {
```
lencioni: Can you expand on what you mean by renamed properties?

If you look at line 284 you'll see `if (!originalElements.hasOwnProperty(elementNames[i])) {` which means that this should only be running on properties that were not there before the `prefixAll()` call but were there after it.

src/generate.js
```diff
@@ -258,81 +270,71 @@ export const generateCSSRuleset = (
...
-                // The value is prefixed. Sort this just before the key with
-                // the unprefixed value.
-                return sortOrder[key] - 0.25;
-            } else {
-                // Either the key and value are unprefixed here, or this is a
-                // prefixed key. Either way, this is handled by the sortOrder
-                // calculation above.
-                return sortOrder[key];
+                if (originalStyle && originalElements.hasOwnProperty(originalStyle)) {
+                    const originalIndex = handledDeclarations.keyOrder.indexOf(originalStyle);
```
lencioni: Yep, my benchmark included some things that are prefixed and some things that are not prefixed. This is faster for both because it gets all of the work done in fewer loops and avoids creating new objects in favor of mutating.

src/generate.js
```diff
@@ -258,81 +270,71 @@ export const generateCSSRuleset = (
...
+    const rules = [];
+    for (let i = 0; i < handledDeclarations.keyOrder.length; i ++) {
+        const key = handledDeclarations.keyOrder[i];
+        const value = prefixedElements[key];
+        if (Array.isArray(value)) {
+            // inline-style-prefixer returns an array when there should be
+            // multiple rules for the same key. Here we flatten to multiple
+            // pairs with the same key.
+            for (let j = 0; j < value.length; j++) {
+                rules.push(transformRule(key, value[j], transformValue));
```
lencioni: I think that's correct. At least in the examples I was looking at, this seemed to be the order we wanted. Although I'm not sure if that is always true or if it will remain stable. It might be worth adding a check here to ensure that the original or unprefixed value is last.

src/inject.js
lencioni: I doubt it--I think I just did this out of habit as I find the early return pattern to be generally more readable. Want me to revert this back?

package.json
lencioni: The dependency is very small. Here's the whole thing:

```js

var canUseDOM = !!(
  typeof window !== 'undefined' &&
  window.document &&
  window.document.createElement
);

module.exports = canUseDOM;
```

I think it is best to prefer dependencies over code duplication, but if you feel strongly about it, I can inline it into this project.

src/prefix.js
lencioni: Whoops!

src/prefix.js
lencioni: Yeah, that's a fair point. I'll look into this when I have a moment.

src/inject.js
lencioni: oops! Done.

src/inject.js
```diff
@@ -99,10 +100,18 @@ const stringHandlers = {
             // Since keyframes need 3 layers of nesting, we use `generateCSS` to
             // build the inner layers and wrap it in `@keyframes` ourselves.
             let finalVal = `@keyframes ${name}{`;
-            Object.keys(val).forEach(key => {
-                finalVal += generateCSS(
-                    key, [val[key]], selectorHandlers, stringHandlers, false);
-            });
+
+            if (val instanceof OrderedElements) {
```
lencioni: I'm not sure if we can change things so that this fork doesn't need to happen. If we can't it might be worth adding a utility function that takes either a plain object or OrderedElements instance and iterates over the keys/values for convenience.

src/ordered-elements.js
```diff
@@ -1,35 +1,61 @@
...
+    constructor() {
+        this.elements = {};
+        this.keyOrder = [];
     }
 
     forEach(callback /* : (string, any) => void */) {
         for (let i = 0; i < this.keyOrder.length; i++) {
-            callback(this.keyOrder[i], this.elements[this.keyOrder[i]]);
+            // (value, key) to match Map's API
+            callback(this.elements[this.keyOrder[i]], this.keyOrder[i]);
```
lencioni: Note: I changed the order of arguments here to match how `Map` works.

tests/generate_test.js
```diff
@@ -134,6 +154,24 @@ ${formatStyles(actual)}
         }], '.foo{color:blue !important;}');
     });
 
+    it('does not mutate nested objects', () => {
```
lencioni: Note: this test exposes a bug that is currently in master that this commit fixes.

lencioni: This might be #226?

lencioni: Also #231

src/generate.js
xymostech: I actually don't remember what I meant by renamed properties.. :P

What I meant in general is that are we sure that the only properties that were not there before the `prefixAll()` call are ones that start with `Webkit`, `Moz`, or `ms`? Could there be others?

package.json
xymostech: Oh, interesting. I guess if it's available in node 4, it's probably fine for us to remove it.

tests/inject_test.js
```diff
@@ -369,11 +369,11 @@ describe('String handlers', () => {
             css(sheet.animate);
             flushToStyleTag();
 
-            assertStylesInclude('@keyframes keyframe_1kmnkfo');
+            assertStylesInclude('@keyframes keyframe_tmjr6');
```
xymostech: why did this change?

src/ordered-elements.js
```diff
@@ -1,35 +1,61 @@
...
         }
     }
 
     set(key /* : string */, value /* : any */) {
         if (!this.elements.hasOwnProperty(key)) {
             this.keyOrder.push(key);
         }
+
```
xymostech: It seems non-intuitive that `.set()` does a deep merge of the elements. Could we maybe pull this deep merge into a separate method, or if this isn't being used anywhere else, rename this?

src/inject.js
xymostech: Nah, that's fine, I was just curious. :)

package.json
xymostech: Nope, using that module seems fine! I'm just cautious about including something if it's trying to do something more complicated.

src/prefix.js
xymostech: :( That's a pretty big size bump. I'd like to get this landed because of the bug fix in it, but I'm worried about throwing this significant change onto users. Could we maybe pull this change out into a separate PR to discuss?

src/inject.js
xymostech: Seems fine for now. Maybe add a TODO for later?

src/generate.js
lencioni: I just looked through the code of inline-style-prefixer and I was unable to find any others. @rofrischmann can you confirm?

src/inject.js
lencioni: Yes, I will add a TODO.

My main concern is less about this specific piece of code and more about what this means for folks who have added their own string handlers that will also need to account for this.

src/ordered-elements.js
lencioni: Yeah this makes sense to me. I'll try to look into this soon.

src/prefix.js
lencioni: Sounds good!

tests/inject_test.js
lencioni: I don't remember exactly anymore, but I think it was because previously the nested objects were plain objects and part of my change (making things actually recursive) caused them to be OrderedElements objects, so they serialized differently. This is related to the code I commented on here: https://github.com/Khan/aphrodite/pull/216/files#r107288563

lencioni:
Once #233 is merged, I'll rebase this again and it will be only 2 commits.

SamyPesse:
Thanks @lencioni 🍻 

Any chance someone from Khan can review it ? (@xymostech) This PR fixes #231.

src/inject.js
```diff
@@ -99,10 +104,23 @@ const stringHandlers = {
...
-                finalVal += generateCSS(
-                    key, [val[key]], selectorHandlers, stringHandlers, false);
-            });
+
+            // TODO see if we can find a way where checking for OrderedElements
+            // here is not necessary. Alternatively, perhaps we should have a
+            // utility method that can iterate over either a plain object, an
+            // instance of OrderedElements, or a Map, and then use that here and
+            // elsewhere.
+            if (val instanceof OrderedElements) {
```
lencioni: This is the bit that I am mostly worried about. But, it looks like string handlers can't be added by extensions, so maybe this is safe?

elischutze:
this looks great I support this

lencioni:
I have @xymostech's [approval for these changes](https://github.com/Khan/aphrodite/pull/216#pullrequestreview-28549298), so I'm going to merge this. But @xymostech or someone else with access will either need to publish a new version or give me access (I'm lencioni on npm).

jlfwong:
@lencioni Just gave you access! The version bump should happen with `npm version`, then the result committed and pushed

lencioni:
Great! I should have some time this afternoon to take care of this.

PR #232, opened 2017-03-23
Exclude tests directory from test coverage, update .gitignore and .eslintignore

lencioni:
I pulled these commits off of #216 to reduce the size of that PR.

We don't need to ensure that tests fully pass coverage. nyc by default
has some patterns to exclude tests, but they didn't match the naming
convention of this project, so I am updating it here.

https://github.com/istanbuljs/nyc#excluding-files

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,3 +1,4 @@
+*/node_modules
+coverage
 dist
 lib
-*/node_modules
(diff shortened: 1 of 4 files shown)


lencioni:
That's odd. Not entirely sure why tests are failing on this PR. It looks like maybe the display flex prefixed styles are in a different order. Thoughts?

lencioni:
I'm going to merge this now since it is so straightforward so I can rebase my other branch on top of it more easily. Please let me know if you'd like any changes!

PR #233, opened 2017-03-23
Optimize css() some

lencioni:
I am pulling these commits out of #216 because I think they are pretty straightforward and should be easier to merge in this way.

### DIFF
--- a/src/generate.js
+++ b/src/generate.js
@@ -203,20 +203,27 @@ const runStringHandlers = (
         return declarations;
     }
 
-    return declarations.map((key, val) => {
-        // If a handler exists for this particular key, let it interpret
-        // that value first before continuing
-        if (stringHandlers.hasOwnProperty(key)) {
+    const stringHandlerKeys = Object.keys(stringHandlers);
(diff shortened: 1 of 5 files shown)


lencioni:
@xymostech this PR should be pretty straightforward and unblocks some of the improvements I wanted to do on #216. Any chance you are able to take a look soon?

lencioni:
Yeah that's right, but I'm a little nervous about my other PR because of
how it affects plugins and how some of the logic feels a bit spread out. It
might be good to take a step back and see if we can do it more cleanly.

I'm out of town until Tuesday so I probably won't be very responsive until
then.

lencioni:
To close the loop a bit, v1.2.1 just went to production for Airbnb, and it reduced our server rendering times by ~15% (v1.2.0: 285ms, v1.2.1: 245ms). Obviously, this will also help on the client side as well, but the numbers are a little more challenging to consistently measure there.

xymostech:
Thank you so much for pushing this forwards, @lencioni! Glad to hear that it improved performance! :)

PR #238, opened 2017-05-05
Fix bug where styles have their media queries removed on IE

reklawnos:
In IE 9, appending two styles with media queries to a `StyleSheet`'s `cssText` property results in the first style's media query being removed and that first style will always apply without the query.

For example:

```js
var styleTag = document.createElement('style');
document.body.appendChild(styleTag);
styleTag.styleSheet.cssText +=
  '@media (min-width: 500px) { body { background-color: blue; } }';
styleTag.styleSheet.cssText +=
  '@media (min-width: 700px) { body { background-color: red; } }';
```

Causes the body element to be red for window widths ≥ 700px, but it is blue otherwise, even at widths < 500px.

This change adds a `styleString` variable that keeps track of the contents of the current `styleTag`, and updates the `cssText` content with `styleString` instead of appending to it with `+=`.

[Here's a JS Bin that demonstrates the issue](http://jsbin.com/nitezecovo/edit), which includes a commented-out solution. Note that the JS Bin won't work outside of IE. If you don't want to open up IE you can take a look at these gifs, recorded on IE 9:

### Using `styleSheet.cssText`
broken example

### Using `innerText`
working example

### DIFF
--- a/src/inject.js
+++ b/src/inject.js
@@ -21,6 +21,12 @@ type ProcessedStyleDefinitions = {
 // faster.
 let styleTag = null;
 
+// The current CSS contents of the style tag. This is an optimization for IE,
+// as using `styleTag.styleSheet.cssText` is significantly faster than altering
+// the text content of the `styleTag`, but `cssText` doesn't retain media queries,
+// causing them to be blown away by subsequent additions.
+let styleString /* : string */ = '';
+
 // Inject a string of styles into a <style> tag in the head of the document. This
 // will automatically create a style tag and then continue to use it for
 // multiple injections. It will also use a style tag with the `data-aphrodite`
@@ -41,13 +47,17 @@ const injectStyleTag = (cssContents /* : string */) => {
             styleTag.type = 'text/css';
             styleTag.setAttribute("data-aphrodite", "");
             head.appendChild(styleTag);
+        } else if (styleTag.styleSheet) {
+            // $FlowFixMe: legacy Internet Explorer compatibility
+            styleString = styleTag.innerText;
         }
     }
 
 
     if (styleTag.styleSheet) {
+        styleString += cssContents;
         // $FlowFixMe: legacy Internet Explorer compatibility
-        styleTag.styleSheet.cssText += cssContents;
+        styleTag.styleSheet.cssText = styleString;
     } else {
         styleTag.appendChild(document.createTextNode(cssContents));
     }
@@ -190,6 +200,7 @@ export const reset = () => {
     alreadyInjected = {};
     isBuffering = false;
     styleTag = null;
+    styleString = '';
 };
 
 export const startBuffering = () => {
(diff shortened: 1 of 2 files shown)


reklawnos:
[clabot:check]

src/inject.js
```diff
@@ -47,7 +47,7 @@ const injectStyleTag = (cssContents /* : string */) => {
 
     if (styleTag.styleSheet) {
         // $FlowFixMe: legacy Internet Explorer compatibility
-        styleTag.styleSheet.cssText += cssContents;
+        styleTag.innerText += cssContents;
```
lencioni: Did you do any profiling to see what the performance impact this might be?

Also, I wonder if we should be using `insertRule()` instead?

https://developer.mozilla.org/en-US/docs/Web/API/CSSStyleSheet/insertRule

reklawnos:
@lencioni I changed the approach a bit so that we're still using `cssText`. `innerText` had a roughly 10x performance penalty, and since AFACT this issue only affects styles with media queries that's not really acceptable.

I found that `insertRule` had around the same performance as the current `cssText` approach in IE, but it appeared to be quite a bit more performant than the `appendChild` approach in Chrome/Firefox. Its performance is pretty consistent, whereas `appendChild` gets quite a bit slower every time new styles are added.

There's a fair amount of work that'd need to be done to get `insertRule` to function. There was some work done to use `insertRule` in #83 and #157, and some discussion in #76. Looks like it would require some refactoring and changes to several tests. `insertRule` also throws when you pass it selectors the browser doesn't recognize, e.g. `.foo:-moz-placeholder` in Chrome. We could probably get some insight from looking at [mattkrick's fork of Aphrodite](https://github.com/mattkrick/aphrodite) which uses `insertRule`.

src/inject.js
```diff
@@ -41,13 +47,17 @@ const injectStyleTag = (cssContents /* : string */) => {
...
+            styleString = styleTag.innerText;
         }
     }
 
 
     if (styleTag.styleSheet) {
+        styleString += cssContents;
         // $FlowFixMe: legacy Internet Explorer compatibility
-        styleTag.styleSheet.cssText += cssContents;
+        styleTag.styleSheet.cssText = styleString;
```
lencioni: I wonder if you could avoid managing `styleString` by doing this instead?

```js
styleTag.styleSheet.cssText = styleTag.innerText + cssContents;
```

With your current approach I mostly worry about something else modifying the stylesheet and then that getting blown away (in addition to the extra bookkeeping complexity). Doesn't seem super likely, but if we can trivially avoid it might as well, right?

reklawnos: I agree that we should avoid having to keep track of `styleString` if possible, but that solution won't quite work because `innerText` doesn't get updated when `cssText` is altered.

lencioni:
Closing, as I believe this was fixed by #240.

PR #240, opened 2017-05-12
Use `insertRule` to inject styles

reklawnos:
Instead of appending a new text node to a style tag or appending to the `cssText` property of the stylesheet (which broke media queries in IE as documented in #238), use `insertRule` to add CSS rules to the `style` element.

This change requires styles to be generated as an array of strings with one rule per string, rather than a single large string that contains multiple rules.

This was (at least provisionally) implemented in  #83 and #157 and discussed in #76, but it didn't seem like those were going anywhere.

### DIFF
--- a/src/generate.js
+++ b/src/generate.js
@@ -14,12 +14,12 @@ const prefixAll = createPrefixer(staticData);
 /* ::
 import type { SheetDefinition } from './index.js';
 type StringHandlers = { [id:string]: Function };
-type SelectorCallback = (selector: string) => any;
+type SelectorCallback = (selector: string) => string[];
 export type SelectorHandler = (
     selector: string,
     baseSelector: string,
     callback: SelectorCallback
-) => string[] | null;
+) => string[] | string | null;
 */
 
 /**
@@ -49,9 +49,10 @@ export type SelectorHandler = (
  *   callback('.foo:nth-child(2n):hover')
  *
  * to generate its subtree `{ color: 'red' }` styles with a
- * '.foo:nth-child(2n):hover' selector. The callback would return CSS like
+ * '.foo:nth-child(2n):hover' selector. The callback would return an array of CSS
+ * rules like
  *
- *   '.foo:nth-child(2n):hover{color:red !important;}'
+ *   ['.foo:nth-child(2n):hover{color:red !important;}']
  *
  * and the handler would then return that resulting CSS.
  *
@@ -67,16 +68,12 @@ export type SelectorHandler = (
  * @param {function} generateSubtreeStyles: A function which can be called to
  *     generate CSS for the subtree of styles corresponding to the selector.
  *     Accepts a new baseSelector to use for generating those styles.
- * @returns {?string[]} The generated CSS for this selector, or null if we don't
- *     handle this selector.
+ * @returns {string[] | string | null} The generated CSS for this selector, or
+ *     null if we don't handle this selector.
  */
-export const defaultSelectorHandlers = [
+export const defaultSelectorHandlers /* : SelectorHandler[] */ = [
     // Handle pseudo-selectors, like :hover and :nth-child(3n)
-    function pseudoSelectors(
-        selector /* : string */,
-        baseSelector /* : string */,
-        generateSubtreeStyles /* : Function */
-    ) /* */ {
+    function pseudoSelectors(selector, baseSelector, generateSubtreeStyles) {
         if (selector[0] !== ":") {
             return null;
         }
@@ -84,17 +81,13 @@ export const defaultSelectorHandlers = [
     },
 
     // Handle media queries (or font-faces)
-    function mediaQueries(
-        selector /* : string */,
-        baseSelector /* : string */,
-        generateSubtreeStyles /* : Function */
-    ) /* */ {
+    function mediaQueries(selector, baseSelector, generateSubtreeStyles) {
         if (selector[0] !== "@") {
             return null;
         }
         // Generate the styles normally, and then wrap them in the media query.
         const generated = generateSubtreeStyles(baseSelector);
-        return [`${selector}{${generated}}`];
+        return [`${selector}{${generated.join('')}}`];
     },
 ];
 
@@ -170,7 +163,17 @@ export const generateCSS = (
             if (result != null) {
                 // If the handler returned something, add it to the generated
                 // CSS and stop looking for another handler.
-                generatedStyles.push(...result);
+                if (Array.isArray(result)) {
+                    generatedStyles.push(...result);
+                } else {
+                    // eslint-disable-next-line
+                    console.warn(
+                        'WARNING: Selector handlers should return an array of rules.' +
+                        'Returning a string containing multiple rules is deprecated.',
+                        handler,
+                    );
+                    generatedStyles.push(`@media all {${result}}`);
+                }
                 return true;
             }
         });
--- a/src/generate.js
+++ b/src/generate.js
@@ -19,7 +19,7 @@ export type SelectorHandler = (
     selector: string,
     baseSelector: string,
     callback: SelectorCallback
-) => string | null;
+) => string[] | null;
 */
 
 /**
@@ -67,7 +67,7 @@ export type SelectorHandler = (
  * @param {function} generateSubtreeStyles: A function which can be called to
  *     generate CSS for the subtree of styles corresponding to the selector.
  *     Accepts a new baseSelector to use for generating those styles.
- * @returns {?string} The generated CSS for this selector, or null if we don't
+ * @returns {?string[]} The generated CSS for this selector, or null if we don't
  *     handle this selector.
  */
 export const defaultSelectorHandlers = [
@@ -94,7 +94,7 @@ export const defaultSelectorHandlers = [
         }
         // Generate the styles normally, and then wrap them in the media query.
         const generated = generateSubtreeStyles(baseSelector);
-        return `${selector}{${generated}}`;
+        return [`${selector}{${generated}}`];
     },
 ];
 
@@ -147,7 +147,7 @@ export const generateCSS = (
     selectorHandlers /* : SelectorHandler[] */,
     stringHandlers /* : StringHandlers */,
     useImportant /* : boolean */
-) /* : string */ => {
+) /* : string[] */ => {
     const merged = new OrderedElements();
 
     for (let i = 0; i < styleTypes.length; i++) {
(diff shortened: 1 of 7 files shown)


src/generate.js
```diff
@@ -94,7 +94,7 @@ export const defaultSelectorHandlers = [
         }
         // Generate the styles normally, and then wrap them in the media query.
         const generated = generateSubtreeStyles(baseSelector);
-        return `${selector}{${generated}}`;
+        return [`${selector}{${generated}}`];
```
lencioni: I wonder if we actually need to create a new array here or if we can continue returning a string and handle it properly elsewhere. Also, do we need to change `pseudoSelectors` above to also return an array?

src/inject.js
```diff
@@ -44,12 +44,21 @@ const injectStyleTag = (cssContents /* : string */) => {
...
 
+    const sheet = ((styleTag.styleSheet || styleTag.sheet /* : any */) /* : CSSStyleSheet */);
 
-    if (styleTag.styleSheet) {
-        // $FlowFixMe: legacy Internet Explorer compatibility
-        styleTag.styleSheet.cssText += cssContents;
+    if (sheet.insertRule) {
+        cssContents.forEach((rule) => {
+            try {
+                sheet.insertRule(rule, sheet.cssRules.length);
```
lencioni: It might be worth moving this out into a variable that we increment, to avoid having to read this value every time.

```js
let numberOfRules = sheet.cssRules.length;
cssContents.forEach((rule) => {
  try {
    sheet.insertRule(rule, numberOfRules);
    numberOfRules += 1;
  } catch(e) { ... }
});
```

src/inject.js
```diff
@@ -44,12 +44,21 @@ const injectStyleTag = (cssContents /* : string */) => {
         }
     }
 
+    const sheet = ((styleTag.styleSheet || styleTag.sheet /* : any */) /* : CSSStyleSheet */);
 
-    if (styleTag.styleSheet) {
-        // $FlowFixMe: legacy Internet Explorer compatibility
-        styleTag.styleSheet.cssText += cssContents;
+    if (sheet.insertRule) {
+        cssContents.forEach((rule) => {
```
lencioni: I wonder if we should make this a regular for loop for perf

src/generate.js
reklawnos: `generateSubtreeStyles` should have the same return type as `generateCSS` so `pseudoSelectors` should already return an array. Would it make sense to make the Flow type for `generateSubtreeStyles` more specific?

We could have `generateCSS` handle a string being returned here, but we never want a selector handler to return strings containing multiple rules, as `insertRule` expects a single rule to be passed. My thought here was that selector handlers should always return an array of strings with one rule per string.

This would be a breaking change to the current API, so perhaps we could, say, wrap strings returned by a selector handler in a `@media all {...}` query and warn about that return type being deprecated. Or just bump by a major version.

lencioni: Ah, I see. The alternative would be to do something like this on line 173:

```js
if (Array.isArray(result)) {
  generatedStyles.push(...result);
} else {
  generatedStyles.push(result);
}
```

I think making the breaking change is okay. @xymostech do you have any thoughts on this?

lencioni: Also, yes I think it would be good to improve the flow types where we can.

src/inject.js
```diff
@@ -44,12 +44,21 @@ const injectStyleTag = (cssContents /* : string */) => {
...
+            } catch(e) {
+                // The selector for this rule wasn't compatible with the browser
+            }
+        });
     } else {
-        styleTag.appendChild(document.createTextNode(cssContents));
+        const currentStyleTag = styleTag;
+        cssContents.forEach((rule) => {
+            currentStyleTag.innerText = (currentStyleTag.innerText || '') + rule;
+        });
```
xymostech: We probably want to batch update this, like `currentStyleTag.innerText = (currentStyleTag.innerText || '') + cssContents.join('');`.

src/inject.js
```diff
@@ -19,17 +19,17 @@ type ProcessedStyleDefinitions = {
 // inserted anything yet. We could find this each time using
 // `document.querySelector("style[data-aphrodite"])`, but holding onto it is
 // faster.
-let styleTag = null;
+let styleTag /* : ?HTMLStyleElement */ = null;
```
xymostech: <3

src/inject.js
```diff
@@ -44,12 +44,21 @@ const injectStyleTag = (cssContents /* : string */) => {
         }
     }
 
+    const sheet = ((styleTag.styleSheet || styleTag.sheet /* : any */) /* : CSSStyleSheet */);
 
-    if (styleTag.styleSheet) {
-        // $FlowFixMe: legacy Internet Explorer compatibility
-        styleTag.styleSheet.cssText += cssContents;
```
xymostech: Could you test this change in IE? Looks like this "IE compatibility" branch got changed.

src/inject.js
```diff
@@ -137,7 +146,7 @@ const stringHandlers = {
 let alreadyInjected = {};
 
 // This is the buffer of styles which have not yet been flushed.
-let injectionBuffer = "";
+let injectionBuffer = [];
```
xymostech: Could you give this a `: string[]` type to help flow make sure that these strings can't be null?

src/generate.js
xymostech: Hmmm. I think the funky `@media all{}` with a warning solution sounds the best to me, but I'm not that attached to it. I don't know how many people are using extensions right now, but I'd prefer if we didn't break it already?

(and as a ping, the `psuedoSelectors` function above needs to be changed too, right?)

src/inject.js
```diff
@@ -19,17 +19,17 @@ type ProcessedStyleDefinitions = {
...
-let styleTag = null;
+let styleTag /* : ?HTMLStyleElement */ = null;
 
 // Inject a string of styles into a <style> tag in the head of the document. This
 // will automatically create a style tag and then continue to use it for
 // multiple injections. It will also use a style tag with the `data-aphrodite`
 // tag on it if that exists in the DOM. This could be used for e.g. reusing the
 // same style tag that server-side rendering inserts.
-const injectStyleTag = (cssContents /* : string */) => {
+const injectStyleTag = (cssContents /* : string[] */) => {
```
xymostech: Could this argument name and docstring be updated?

src/inject.js
reklawnos: `insertRule` is supported on ≥ IE 9. I tested it on IE 9 and AFAICT it worked just fine. What browsers does Aphrodite guarantee support for? The `innerText` fallback should work for browsers without `insertRule`, but I just want to get idea of how far back I should test this.

xymostech: Testing in IE 9 sounds good enough to me! Thanks!

src/inject.js
reklawnos: Ran both versions through your benchmark several times. There was no noticeable difference between the `forEach` version and the `for` loop version, but a small speedup when storing `sheet.cssRules.length` in a variable.

reklawnos:
@lencioni @xymostech PTAL, addressed comments and updated to use the `@media all` solution I suggested.

src/inject.js
xymostech: ping on this?

tests/generate_test.js
```diff
@@ -317,7 +318,24 @@ ${formatStyles(actual)}
...
+            const generatedBefore = callback(baseSelector + '::before');
+            const generatedAfter = callback(baseSelector + '::after');
+            return `${generatedBefore} ${generatedAfter}`;
+        };
+
+        assertCSS('.foo', [{
+            '^': {
+                color: 'red',
+            },
+        }], ['@media all {.foo::before{color:red;} .foo::after{color:red;}}'], [handler], {}, false);
```
xymostech: <3

lencioni:
I don't think I'll be able to look for a while so don't hold it up for me.

Trolleymusic:
Hey @reklawnos @xymostech, any plans to merge this soon?

tests/inject_test.js
```diff
@@ -462,9 +464,9 @@ describe('String handlers', () => {
             flushToStyleTag();
 
             assertStylesInclude('@keyframes keyframe_d35t13');
-            assertStylesInclude('0%{opacity:0;-webkit-transform:scale(0.75) translate3d(1px, 2px, 0);-ms-transform:scale(0.75) translate3d(1px, 2px, 0);transform:scale(0.75) translate3d(1px, 2px, 0);}');
-            assertStylesInclude('100%{opacity:1;-webkit-transform:scale(1) translate3d(1px, 2px, 0);-ms-transform:scale(1) translate3d(1px, 2px, 0);transform:scale(1) translate3d(1px, 2px, 0);}');
-            assertStylesInclude('animation-name:keyframe_d35t13');
+            assertStylesInclude('0% {opacity: 0; -webkit-transform: scale(0.75) translate3d(1px,2px,0); -ms-transform: scale(0.75) translate3d(1px,2px,0); transform: scale(0.75) translate3d(1px,2px,0);}');
```
lencioni: Looks like some of the spaces you removed here are still there in the test

lencioni:
I think this looks good. There is one test that has a minor failure on it. I'll follow up with a fix for that after merging this. Thanks for your work on this!

src/generate.js
```diff
@@ -170,7 +163,17 @@ export const generateCSS = (
...
+                if (Array.isArray(result)) {
+                    generatedStyles.push(...result);
+                } else {
+                    // eslint-disable-next-line
+                    console.warn(
+                        'WARNING: Selector handlers should return an array of rules.' +
+                        'Returning a string containing multiple rules is deprecated.',
+                        handler,
+                    );
+                    generatedStyles.push(`@media all {${result}}`);
```
lencioni: I think this may be a semver major change, since it could theoretically cause some styles to stop working in browsers that don't support media queries. For IE, this applies to IE8 and older.

Practically speaking, however, I would be surprised if this actually breaks anything for anyone, so I am tempted to go with semver minor.

What do you think @xymostech?

calinoracation: Just my 2c, but if I understand correctly, if someone _was not_ using media queries before this change, and needed to support IE8 for some reason, then it seems like this should likely be considered a breaking change, as all styles would no longer work? 

Is there a concern that bumping semver major would cause an unnecessary burden to consumers?

lencioni: Yeah I think you are correct. My main concern is that I don't want to do a major release if there is no practical benefit to it, but I guess there's no way to know so probably best to stay on the safe side.

calinoracation: Wondering if opening a new PR to bump the version of `inline-style-prefixer` https://github.com/rofrischmann/inline-style-prefixer/blob/master/Changelog.md, which contains a few bug fixes and reported 30% performance improvement since , along with noting some potential internal perf benefits would be of practical benefit?

lencioni: Yes that sounds great to me!

PR #242, opened 2017-05-31
Add missing paren to example in README.md

omerzach:


xymostech:
Thank you!

omerzach:
[clabot:check]
But I guess that wasn't really necessary ;)

xymostech:
Yeah, we don't usually require cla signatures for docs changes, but the bot doesn't know that. :) Thanks for the contribution!

PR #244, opened 2017-06-06
RunKit example: simplify output

pouwerkerk:
Collapse `console.log`s via a template literal so that the HTML with CSS is rendered in the RunKit result. RunKit behaves like a REPL so the last value is returned.

The existing [RunKit example](https://npm.runkit.com/aphrodite) is great, so this just takes it slightly further so you can see how your browser would render the styles that Aphrodite generates.

You can see it in action here: https://runkit.com/pieter/aphrodite-runkit-example

### Before
screen shot 2017-06-06 at 11 34 07 am

### After
screen shot 2017-06-06 at 11 34 47 am

### DIFF
--- a/examples/runkit.js
+++ b/examples/runkit.js
@@ -29,5 +29,4 @@ const {css, html} = StyleSheetServer.renderStatic(() => {
 });
 
 // Observe our output HTML and the Aphrodite-generated CSS
-console.log("Output HTML:", html);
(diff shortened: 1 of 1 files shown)


pouwerkerk:
[clabot:check]

lencioni:
This looks great to me. Thanks so much for the contribution!

PR #246, opened 2017-06-12
Add compile-time option for class minification

gilbox:
address this existing todo:
> // TODO(emily): Make a 'production' mode which doesn't prepend
> // the class name here, to make the generated CSS smaller.

@lencioni

### DIFF
--- a/README.md
+++ b/README.md
@@ -596,8 +596,8 @@ It isn't determinate whether divs will be red or blue.
 
 ## Minify class names
 
-Minify class names by setting the environment variable `process.env.APHRODITE_KEYS` 
-to the string value `MINIFIED`. 
+Minify class names by setting the environment variable `process.env.NODE_ENV` 
+to the string value `production`. 
 
 # Tools
 
--- a/README.md
+++ b/README.md
@@ -594,6 +594,11 @@ css(styles2.globals);
 
 It isn't determinate whether divs will be red or blue.
 
+## Minify class names
+
+Minify class names by setting the environment variable `process.env.APHRODITE_KEYS` 
+to the string value `MINIFIED`. 
+
 # Tools
 
 - [Aphrodite output tool](https://output.jsbin.com/qoseye) - Paste what you pass to `StyleSheet.create` and see the generated CSS
--- a/src/exports.js
+++ b/src/exports.js
@@ -21,9 +21,8 @@ const StyleSheet = {
     create(sheetDefinition /* : SheetDefinition */) {
         return mapObj(sheetDefinition, ([key, val]) => {
             return [key, {
-                // TODO(emily): Make a 'production' mode which doesn't prepend
-                // the class name here, to make the generated CSS smaller.
-                _name: `${key}_${hashObject(val)}`,
+                _name: process.env.APHRODITE_KEYS !== 'MINIFIED' ?
+                    `${key}_${hashObject(val)}` : `_${hashObject(val)}`,
                 _definition: val
             }];
         });
--- a/src/exports.js
+++ b/src/exports.js
@@ -21,6 +21,7 @@ const StyleSheet = {
     create(sheetDefinition /* : SheetDefinition */) {
         return mapObj(sheetDefinition, ([key, val]) => {
             return [key, {
+                // TODO(gil): Further minify the -O_o--combined hashes
                 _name: process.env.APHRODITE_KEYS !== 'MINIFIED' ?
                     `${key}_${hashObject(val)}` : `_${hashObject(val)}`,
                 _definition: val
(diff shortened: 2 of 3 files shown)


gilbox:
[clabot:check]

README.md
```diff
@@ -594,6 +594,11 @@ css(styles2.globals);
 
 It isn't determinate whether divs will be red or blue.
 
+## Minify class names
+
+Minify class names by setting the environment variable `process.env.APHRODITE_KEYS`
```
xymostech: Is there a reason we want to add an additional env variable, instead of just using `process.env.NODE_ENV === "production"`?

src/exports.js
```diff
@@ -21,9 +21,8 @@ const StyleSheet = {
     create(sheetDefinition /* : SheetDefinition */) {
         return mapObj(sheetDefinition, ([key, val]) => {
             return [key, {
-                // TODO(emily): Make a 'production' mode which doesn't prepend
-                // the class name here, to make the generated CSS smaller.
-                _name: `${key}_${hashObject(val)}`,
+                _name: process.env.APHRODITE_KEYS !== 'MINIFIED' ?
```
xymostech: Another thing that we were considering doing is making something like `css(styles.a, styles.b)` generate only a single hash class name, like `_abcdef` instead of combining two hashes like `_abcdef-O_o-_012345`. Not sure if you'd like to try tackling that here as well, but leaving a TODO about it would be nice if not. :)

If we do want to tackle it, we should figure out whether we need to make our hash longer. Currently, our hash is 6 base-36 characters, which means we'd expect collisions once there's ~47k styles on the page. Is that enough? On the largest KA pages, we end up seeing ~500 independent styles, but our website isn't very SPA-y, so the stylesheet gets cleared out pretty often. If we add uppercase letters, that bumps us up to ~240k styles on a page before seeing a collision.

README.md
gilbox: The reason I added another env variable is because it seems reasonable that someone might not want to minimize class names in production in order to make debugging production-only problems easier.

src/exports.js
gilbox: I was hoping to combine hashes in a separate PR. I'll add the TODO for now.

xymostech: Okay, great! Doing it separately sounds good :)

README.md
lencioni: I'm a little hesitant to require that folks use a new env variable in order to benefit from this.

gilbox: no problem, i'll change it to use NODE_ENV

xymostech:
Thanks! Looking forward to your other PR :)

jlfwong:
One potential issue with using this optimization is that it dramatically increases the probability of hash collision. I believe when we were originally considering doing this, we planned on using a longer hash when omitting the class name.

src/exports.js
```diff
@@ -21,9 +21,9 @@ const StyleSheet = {
     create(sheetDefinition /* : SheetDefinition */) {
         return mapObj(sheetDefinition, ([key, val]) => {
             return [key, {
-                // TODO(emily): Make a 'production' mode which doesn't prepend
-                // the class name here, to make the generated CSS smaller.
-                _name: `${key}_${hashObject(val)}`,
+                // TODO(gil): Further minify the -O_o--combined hashes
+                _name: process.env.NODE_ENV === 'production' ?
+                    `_${hashObject(val)}` : `${key}_${hashObject(val)}`,
```
lencioni: Any reason to keep the leading `_` here?

gilbox: class names can't start w/digits

xymostech:
@jlfwong In this PR, we're still doing the `-O_o-` concatenation of class names, so there still isn't a huge chance of collisions because you'd need a very large number of actually unique style objects, instead of just a large number of combinations. It sounds like @gilbox is going to do the further step of hashing the final class names down to a single hash, so we can discuss using a longer hash there?

luixxiul:
Is it possible to avoid minifying even if the variable `process.env.NODE_ENV` is set to `production`? 

https://github.com/brave/browser-laptop/issues/10029#issuecomment-317886257

gilbox:
@luixxiul when [this was discussed](https://github.com/Khan/aphrodite/pull/246#discussion_r121533492) we didn't have a specific use-case in mind. Maybe open a PR to add a new environment variable that turns off minification?

luixxiul:
That would be great. IMHO the info that the option is enabled by default should have been added to CHANGELOG.md.

gilbox:
Yeah, unfortunately the CHANGELOG.md is a couple of versions out of date.

luixxiul:
I opened a new issue for that: https://github.com/Khan/aphrodite/issues/261

PR #247, opened 2017-06-13
[createClassUID] Minify combined classNames

gilbox:
I'll add specs if there are no major concerns with this approach. The minified classNames for production are generated sequentially (base 36) instead of using a hash. The first class name is always `_1`, then `_2` and so on...

@lencioni 
@xymostech 
@jlfwong

### DIFF
--- a/src/inject.js
+++ b/src/inject.js
@@ -136,6 +136,9 @@ const stringHandlers = {
 // set of class names)
 let alreadyInjected = {};
 
+// The number of classes that have been injected
+let classUIDCount = 0;
+
 // This is the buffer of styles which have not yet been flushed.
 let injectionBuffer = "";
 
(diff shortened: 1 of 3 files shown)


src/inject.js
```diff
@@ -136,6 +136,9 @@ const stringHandlers = {
 // set of class names)
 let alreadyInjected = {};
 
+// The number of classes that have been injected
+let classUIDCount = 0;
```
xymostech: The reason we use hashing in the first place is to ensure that rehydration will work: If we take an approach like this one (which we were using in the original implementation, see https://github.com/Khan/aphrodite/commit/f45f87d25f53644a360979479bf0c0b8a8ba1ff8), then the server and client might disagree on what the number should be for a given style (either because the code ran in a different order on the client or server, or because this is a file-local variable and the server keeps it around in between runs).

So, we definitely need to do hashing in here, an incrementing value isn't going to work. :(

gilbox: Makes sense. What do you think about hashing on the combine className?

ie this value:

```
processedStyleDefinitions.classNameBits.join("-o_O-")
```

xymostech: Yeah, I think that hashing that combined class name is the best approach here. (you probably can just join it, I don't think that adding in some random `-o_O-`s is improving our hash value)

gilbox:
continued in https://github.com/Khan/aphrodite/pull/248/files

PR #248, opened 2017-06-13
Minify combined class names

gilbox:
Hashes the combined class name to create a minified combined class name.

@lencioni
@xymostech
@jlfwong

### DIFF
--- a/src/exports.js
+++ b/src/exports.js
@@ -22,7 +22,7 @@ const StyleSheet = {
         return mapObj(sheetDefinition, ([key, val]) => {
             return [key, {
                 _name: process.env.NODE_ENV === 'production' ?
-                    `_${hashObject(val)}` : `${key}_${hashObject(val)}`,
+                    hashObject(val) : `${key}_${hashObject(val)}`,
                 _definition: val
             }];
         });
--- a/src/exports.js
+++ b/src/exports.js
@@ -21,7 +21,6 @@ const StyleSheet = {
     create(sheetDefinition /* : SheetDefinition */) {
         return mapObj(sheetDefinition, ([key, val]) => {
             return [key, {
-                // TODO(gil): Further minify the -O_o--combined hashes
                 _name: process.env.NODE_ENV === 'production' ?
                     `_${hashObject(val)}` : `${key}_${hashObject(val)}`,
                 _definition: val
--- a/src/exports.js
+++ b/src/exports.js
@@ -1,5 +1,5 @@
 /* @flow */
-import {mapObj, hashObject} from './util';
+import {mapObj, hashString} from './util';
 import {
     injectAndGetClassName,
     reset, startBuffering, flushToString,
@@ -20,9 +20,11 @@ export type MaybeSheetDefinition = SheetDefinition | false | null | void
 const StyleSheet = {
     create(sheetDefinition /* : SheetDefinition */) {
         return mapObj(sheetDefinition, ([key, val]) => {
+            const stringVal = JSON.stringify(val);
             return [key, {
+                _len: stringVal.length,
                 _name: process.env.NODE_ENV === 'production' ?
-                    hashObject(val) : `${key}_${hashObject(val)}`,
+                    hashString(stringVal) : `${key}_${hashString(stringVal)}`,
                 _definition: val
             }];
         });
--- a/src/inject.js
+++ b/src/inject.js
@@ -243,6 +243,12 @@ const processStyleDefinitions = (
     }
 };
 
+// Sum up the lengths of the stringified style definitions (which was saved as _len property)
+// and use modulus to return a single byte hash value.
+const getStyleDefinitionsLengthHash = (styleDefinitions /* : any[] */) /* : string */ => (
+    styleDefinitions.reduce((length, styleDefinition) => length + styleDefinition._len, 0) % 36
+).toString(36);
+
 /**
  * Inject styles associated with the passed style definition objects, and return
  * an associated CSS class name.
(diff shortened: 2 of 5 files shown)


src/exports.js
```diff
@@ -21,7 +21,6 @@ const StyleSheet = {
     create(sheetDefinition /* : SheetDefinition */) {
         return mapObj(sheetDefinition, ([key, val]) => {
             return [key, {
-                // TODO(gil): Further minify the -O_o--combined hashes
                 _name: process.env.NODE_ENV === 'production' ?
                     `_${hashObject(val)}` : `${key}_${hashObject(val)}`,
```
lencioni: Since this is going to be joined and hashed again, we can probably remove the leading underscore here now, right?

gilbox:
ping @lencioni addressed your comment and added other improvements

gilbox:
ping @xymostech

csilvers:
If you need more than 32 bits of hash, you're kinda stuck, since js doesn't provide 64-bit integer types.

You could go back to using a google hash: https://github.com/lovell/farmhash e.g.  It has fingerprint64() which gives you a 64-bit hash (as a string), and is platform independent.  I don't know how fast it is, though.

Your only other solution would be to generate two separate 32-bit hashes and concatenate them.  The most straightforward way to do this, if you're always hashing the same kind of object, is to add a 'salt' field to the object, and hash it once with salt=1 and a second time with salt=2 (or whatever).

Of course, this will double your hashing time.  farmhash is probably faster. :-)

csilvers:
(Blah, maybe not possible -- it looks like farmhash is node-specific, and just a wrapper around some C code.)

src/inject.js
```diff
@@ -269,7 +269,15 @@ export const injectAndGetClassName = (
     if (processedStyleDefinitions.classNameBits.length === 0) {
         return "";
     }
-    const className = processedStyleDefinitions.classNameBits.join("-o_O-");
+
+    let className;
+    if (process.env.NODE_ENV === 'production') {
+        className = processedStyleDefinitions.classNameBits.length === 1 ?
+            `_${processedStyleDefinitions.classNameBits[0]}` :
+            `_${hashString(processedStyleDefinitions.classNameBits.join())}`;
```
lencioni: What if we append `processedStyleDefinitions.classNameBits.length`? That would help reduce the likelihood of collisions.

gilbox:
ping @xymostech @lencioni 
I implemented something similar to @lencioni's suggestion but instead of using the length of `processedStyleDefinitions.classNameBits.length` I used the combined length of the stringified style definition objects since they probably exhibit more variation. However, that length is modded by `36` so that there is only one additional byte added.

src/inject.js
```diff
@@ -243,6 +243,12 @@ const processStyleDefinitions = (
     }
 };
 
+// Sum up the lengths of the stringified style definitions (which was saved as _len property)
+// and use modulus to return a single byte hash value.
```
lencioni: This comment would be 💯  if it explained why we are adding the length to the hash.

src/inject.js
```diff
@@ -269,7 +275,16 @@ export const injectAndGetClassName = (
     if (processedStyleDefinitions.classNameBits.length === 0) {
         return "";
     }
-    const className = processedStyleDefinitions.classNameBits.join("-o_O-");
+
+    let className;
+    if (process.env.NODE_ENV === 'production') {
+        className = processedStyleDefinitions.classNameBits.length === 1 ?
+            `_${processedStyleDefinitions.classNameBits[0]}` :
```
xymostech: should we maybe be adding the length of the style definitions when there's only one class name, too? If only to make the hashes the same size/consistent?

xymostech:
I think the travis failing was a fluke, I restarted it

src/inject.js
gilbox: I'm not sure what the advantage of that would be. Also, before this PR the hashes were different lengths--for example, from our homepage:

```
image_ay4wjb-o_O-background_1h6n1zu-o_O-fadeIn_3jddj2-o_O-backgroundSize_cover_176vplr
```
--so I imagine they still won't be the same size.

gilbox: oh, the other thing is that we'd be hashing a hash unnecessarily. is the additional expense justified?

src/inject.js
gilbox: https://github.com/Khan/aphrodite/pull/248/commits/263b35f98e512a8af345d02bd34d9894e0e3c655

src/inject.js
xymostech: We wouldn't be hashing anything new, just adding the extra length char on, right? You're right though, this probably isn't important. :)

gilbox: oh, you're right double-hashing shouldn't be a thing

xymostech:
Okay, I'm gonna merge this! Sorry for the delay! Maybe once this lands, @lencioni or I could try updating aphrodite on our sites and see if anything breaks and how much smaller the styles get.

lencioni:
That sounds good. I actually work with @gilbox so I'll let him do the honors on our site and report back here with anything interesting. 💃

xymostech:
Oh ha, I didn't realize! Sounds wonderful :)

gilbox:
@xymostech could we get a release please 🙏

xymostech:
@gilbox Did you test things out on your site? I can make a release but I'd love for someone to do some real testing to make sure this doesn't horribly break things...

gilbox:
ah ok I didn't realize that's what you meant. I'll give it a whirl

gilbox:
I tested with #263, should be good

gilbox:
ping @xymostech for the new version

PR #256, opened 2017-07-14
Make String Handlers Use useImportant

jeresig:
The string handlers were always inserting rules with no `!important`, rather than respecting the settings specified by the library. This updates the string handlers to receive an additional `useImportant` argument, which is then passed down to the various places that need it (rather than just passing `false`). I updated the existing test, which was confirming the wrong input, and also added some more tests to `no-important_test.js` to confirm that it was still being disabled.

In practice this helps to fix the issue where Firefox makes it impossible to override CSS that has `!important` with an animation, unless the animation is, also, using `!important`:
http://tosbourn.com/firefox-honours-important-in-css-animations-no-one-else-seems-to/

Test Plan:
I ran `npm test` and it passed.

### DIFF
--- a/tests/no-important_test.js
+++ b/tests/no-important_test.js
@@ -6,7 +6,7 @@ import {
   StyleSheet,
   css
 } from '../src/no-important.js';
-import { reset } from '../src/inject.js';
+import { reset, startBuffering, flushToStyleTag } from '../src/inject.js';
 
 describe('css', () => {
     beforeEach(() => {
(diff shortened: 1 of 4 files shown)


tests/no-important_test.js
```diff
@@ -39,3 +39,117 @@ describe('css', () => {
...
+                },
+            });
+
+            startBuffering();
+            css(sheet.base);
+            flushToStyleTag();
+
+            assertStylesInclude('font-family:"CoolFont",sans-serif');
+            assertStylesInclude('font-family:CoolFont;');
+            assertStylesInclude("src:url('coolfont.ttf');");
```
xymostech: Great tests!

PR #258, opened 2017-07-18
Updated Typescript definitions and included in package

upta:
It seems that https://github.com/DefinitelyTyped/DefinitelyTyped/pull/12845 kinda died on the vine.

So, I took the .d.ts file from that PR, tweaked it a little bit by:
*  adding an overload to "create" (makes intellisense a bit happier in the default case)
*  changing the parameter type for "css" to model after the Flow type (instead of just "any")

Since you can just include a .d.ts file in your package.json and Typescript will pick it up, I figured that would be the lowest friction distribution method, if you're cool with that.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -103,7 +103,3 @@ declare module "aphrodite" {
         extend( extensions: Extension[] ): Exports;
     }
 }
-
-declare module "aphrodite/no-important" {
-    export * from "aphrodite";
-}
--- a/no-important.d.ts
+++ b/no-important.d.ts
@@ -1,0 +1,1 @@
+export * from "aphrodite";
--- a/package.json
+++ b/package.json
@@ -8,6 +8,7 @@
     "inline-styles"
   ],
   "main": "lib/index.js",
+  "types": "types/index.d.ts",
   "scripts": {
     "lint": "eslint --fix --cache . && flow check",
     "test": "npm run coverage",
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
     "inline-styles"
   ],
   "main": "lib/index.js",
-  "types": "types/index.d.ts",
+  "types": "index.d.ts",
   "scripts": {
     "lint": "eslint --fix --cache . && flow check",
     "test": "npm run coverage",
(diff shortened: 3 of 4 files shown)


upta:
[clabot:check]

abruzzihraig:
Hi @upta , I've tried your declaration file, but it doesn't work for `aphrodite/no-important`. If I use it like `import { StyleSheet, css } from 'aphrodite/no-important';`, there would be an error `Could not find a declaration file for moudle 'aphrodite/no-important'`

upta:
I'll take a look at it.  I never tried it with the no-import since it mostly wasn't my file to start with, just tweaked a few things and added it to the package.json :)

upta:
@abruzzihraig I updated the PR, no-important was working for me, let me know if it behaves better for you now.  Thanks!

abruzzihraig:
@upta thanks bro, it works fine!

screendriver:
Any plans to merge this PR?

index.d.ts
```diff
@@ -0,0 +1,105 @@
+// Type definitions for Aphrodite 1.2.3
+// Project: https://github.com/Khan/aphrodite
+
+declare module "aphrodite" {
+    import * as React from "react";
```
jlfwong: I've been using TypeScript & Aphrodite for a side project, and had to use my own copy because I'm using Aphrodite but not react -- can you remove this, since Aphrodite doesn't have a dependency on React? As an initial pass, it should be fine to just replace `React.CSSProperties` with `any`.

jlfwong:
@upta Let me know if you'd still like to merge this and get credit, otherwise I'm happy to take this over and get it landed

najamkhn:
I get the following error when I'm trying to using `aphrodite`:

```'css' is declared but its value is never read.```

Is this pull request fixing it as well?

upta:
@jlfwong Hey, sorry, I totally forgot you had pinged me until today's notification came in ;)  Feel free to take over, I was just experimenting with aphrodite at the time and figured I'd see if I could help out

lencioni:
I believe this has been accomplished by #302

PR #263, opened 2017-08-02
Fix injectAndGetClassName and ignore null values in styleDefinitions

gilbox:
Fixes regression introduced by #248 
@xymostech


gilbox:
ping @xymostech

lencioni:
Thanks for the fix and the tests!

gilbox:
Ping @xymostech please publish v2.0

lencioni:
I published this as v1.2.4. Thanks for the patience!

PR #277, opened 2017-10-10
Add tests for keyframes with multiple properties

lencioni:
I was investigating a bug and wanted to have concrete proof that it was
not present in Aphrodite, so I wrote this test. Since I took the time to
write it, I figured I might as well commit it up.


lencioni:
I handed it off to someone else, and I'm not sure if they have found the root cause yet. It seems likely related to some recent RTL work we've been doing here https://github.com/airbnb/react-with-styles-interface-aphrodite#built-in-rtl-support

PR #279, opened 2017-10-11
Allow overridden styles to re-order in generated CSS

lencioni:
We ran into a weird scenario, where if you passed the following two
objects to css() in this order:

```js
{
  '@media all and (min-width: 1128px)': {},
},
{
  '@media all and (min-width: 744px)': {
    backgroundColor: 'red',
  },
  '@media all and (min-width: 1128px)': {
    backgroundColor: 'blue',
  },
}
```

you would normally expect it to produce a style that has the 744px
min-width media query first. Unfortunately, because the first object
already had that media query as an ordered key in its OrderedElements
object, it maintained its original position. This caused the 1128px
media query to be output first, which is unexpected and ended up causing
a visual bug.

To fix this, I modified OrderedElements to always move overridden keys
to the end. This now behaves more like you would expect regular CSS to
work.

I needed to add a special case for string handlers, since they aren't
actually overriding a style but rather replacing a property with a newly
computed value.

cc @DanielGarcia-Carrillo

### DIFF
--- a/src/generate.js
+++ b/src/generate.js
@@ -199,9 +199,9 @@ const runStringHandlers = (
     declarations /* : OrderedElements */,
     stringHandlers /* : StringHandlers */,
     selectorHandlers /* : SelectorHandler[] */
-) /* : OrderedElements */ => {
+) /* : void */ => {
     if (!stringHandlers) {
-        return declarations;
+        return;
     }
 
     const stringHandlerKeys = Object.keys(stringHandlers);
@@ -223,8 +223,6 @@ const runStringHandlers = (
             );
         }
     }
-
-    return declarations;
 };
 
 
--- a/src/generate.js
+++ b/src/generate.js
@@ -177,7 +177,7 @@ export const generateCSS = (
         // If none of the handlers handled it, add it to the list of plain
         // style declarations.
         if (!foundHandler) {
-            plainDeclarations.set(key, val);
+            plainDeclarations.set(key, val, true);
         }
     });
 
@@ -224,7 +224,7 @@ const runStringHandlers = (
                 // Preserve order here, since we are really replacing an
                 // unprocessed style with a processed style, not overriding an
                 // earlier style
-                true
+                false
             );
         }
     }
--- a/src/generate.js
+++ b/src/generate.js
@@ -219,7 +219,12 @@ const runStringHandlers = (
             // handlers are very specialized and do complex things.
             declarations.set(
                 key,
-                stringHandlers[key](declarations.get(key), selectorHandlers)
+                stringHandlers[key](declarations.get(key), selectorHandlers),
+
+                // Preserve order here, since we are really replacing an
+                // unprocessed style with a processed style, not overriding an
+                // earlier style
+                true
             );
         }
     }
--- a/src/ordered-elements.js
+++ b/src/ordered-elements.js
@@ -19,10 +19,10 @@ export default class OrderedElements {
         }
     }
 
-    set(key /* : string */, value /* : any */, preserveOrder /* : ?boolean */) {
+    set(key /* : string */, value /* : any */, shouldReorder /* : ?boolean */) {
         if (!this.elements.hasOwnProperty(key)) {
             this.keyOrder.push(key);
-        } else if (!preserveOrder) {
+        } else if (shouldReorder) {
             const index = this.keyOrder.indexOf(key);
             this.keyOrder.splice(index, 1);
             this.keyOrder.push(key);
@@ -40,7 +40,7 @@ export default class OrderedElements {
                 ? this.elements[key]
                 : new OrderedElements();
             value.forEach((value, key) => {
-                nested.set(key, value);
+                nested.set(key, value, shouldReorder);
             });
             this.elements[key] = nested;
             return;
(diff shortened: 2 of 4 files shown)


src/ordered-elements.js
```diff
@@ -19,9 +19,12 @@ export default class OrderedElements {
         }
     }
 
-    set(key /* : string */, value /* : any */) {
+    set(key /* : string */, value /* : any */, preserveOrder /* : boolean */) {
```
xymostech: Hmm, could we maybe have `preserveOrder` default to `true` and set it to `false` in all of the places we explicitly want it to override?

lencioni: We want this to be false in every place except one: string handlers. I'm happy to flip it to be `reorder` or `shouldReorder` and make sure that it is always passed in though if you think that is better.

src/ordered-elements.js
```diff
@@ -19,9 +19,12 @@ export default class OrderedElements {
...
         if (!this.elements.hasOwnProperty(key)) {
             this.keyOrder.push(key);
+        } else if (!preserveOrder) {
+            const index = this.keyOrder.indexOf(key);
+            this.keyOrder.push(this.keyOrder.splice(index, 1)[0]);
```
xymostech: I think splitting this into two lines would be clearer
```
this.keyOrder.splice(index, 1);
this.keyOrder.push(key);
```

tests/ordered-elements_test.js
```diff
@@ -53,6 +53,38 @@ describe("OrderedElements", () => {
...
+            keyOrder: ["b", "a"],
+        }, elems);
+    });
+
+    it('can preserve order when setting', () => {
+        const elems = new OrderedElements();
+
+        elems.set("a", 1);
+        elems.set("b", 1);
+        elems.set("a", 2, { preserveOrder: true });
```
xymostech: this isn't how the API works, right? It should just be a `true` here

src/ordered-elements.js
lencioni: Ah, good call.

tests/ordered-elements_test.js
lencioni: Oh, right. This was an earlier revision. I'll fix now.

src/ordered-elements.js
xymostech: oops, argh, I meant to put this into a review. Other comments incoming...

src/ordered-elements.js
```diff
@@ -19,9 +19,13 @@ export default class OrderedElements {
         }
     }
 
-    set(key /* : string */, value /* : any */) {
+    set(key /* : string */, value /* : any */, preserveOrder /* : ?boolean */) {
```
xymostech: Could we make `preserveOrder` default to `true` and then explicitly set it to `false` in the places where we're overriding? I think it makes more sense for something called `OrderedElements` to preserve order by default.

src/generate.js
```diff
@@ -199,9 +199,9 @@ const runStringHandlers = (
     declarations /* : OrderedElements */,
     stringHandlers /* : StringHandlers */,
     selectorHandlers /* : SelectorHandler[] */
-) /* : OrderedElements */ => {
+) /* : void */ => {
```
xymostech: :+1:

tests/generate_test.js
```diff
@@ -258,6 +258,34 @@ ${formatStyles(actual)}
...
+                padding:20px !important;
+            }
+        }
+        @media (min-width: 400px){
+            .foo{
+                padding:30px !important;
+            }
+        }`, defaultSelectorHandlers);
+    });
+
```
xymostech: do we have a test to make sure string handlers still work correctly?

src/ordered-elements.js
lencioni: Alright, I added a commit that inverts this!

tests/generate_test.js
lencioni: There are some tests for string handlers. Is there a specific test you think I should add?

xymostech: Just to make sure that string handlers don't end up at the end of style declarations. I'm less worried about this now that the argument is inverted, though. :)

lencioni: Yep, there is a test that verifies that already. That's actually how I ended up making this an argument in the first place since I broke that test in my initial pass.

lencioni:
Thanks for the quick review!

xymostech:
You're welcome! Thanks for fixing bugs! :heart:

Soreine:
@lencioni I have run into this behavior just now. Let me state a related question:

I am making a component `<Custom />` that uses a component `<Base />` like so:

```jsx
class Base extends React.Component {
    props: {
        styles: SheetDefinitions
    };

    render() {
        return (
            <div
                className={css(
                    {
                        '@media x': {
                            A
                        },
                        '@media y': {
                            B
                        }
                    },
                    this.props.styles
                )}
            />
        );
    }
}

class Custom extends React.Component {
    render() {
        return (
            <Base
                styles={{
                    '@media x': {
                        C
                    }
                }}
            />
        );
    }
}
```

I would like the generated CSS to be in this order (not to merge the media queries):

```css
@media x {
  A
}

@media y {
  B
}

@media x {
  C
}
```

Instead, I get:

```css
@media y {
  B
}

@media x {
  A,
  C
}
```

How does one avoid the merging of similar media queries ?

Issue #278, opened 2017-10-11
Bundle size badge


vladnicula:
Styled components and glamorous have the bundle size and gziped size badge. 

I'm checking the bundle size differences now on my project. Would have been great if I saw it on the readme just like the other css in js frameworks.

lencioni:
I think this is a great idea and this is related to something I've been thinking about doing here lately as well to reduce the bundle size impact of Aphrodite. Would you be interested in submitting a PR to add the badge?

vladnicula:
https://github.com/Khan/aphrodite/pull/280

lencioni:
Closed by #280

PR #280, opened 2017-10-12
Added gzip and bundle sizes in readme #278

vladnicula:


vladnicula:
[clabot:check]

README.md
```diff
@@ -620,3 +626,7 @@ to the string value `production`.
 Copyright (c) 2016 Khan Academy
 
 [webcomponents]: http://w3c.github.io/webcomponents/spec/custom
+
+[gzip-badge]: http://img.badgesize.io/https://unpkg.com/aphrodite/dist/aphrodite.umd.min.js?compression=gzip&label=gzip%20size
+[size-badge]: http://img.badgesize.io/https://unpkg.com/aphrodite/dist/aphrodite.umd.min.js?label=size
```
lencioni: I'm not entirely sure this will be representative of the bundle size impact for most people. For instance, it looks like this file shows ~24 KiB, but in practice aphrodite is closer to 13 KiB in my bundle.

screen shot 2017-10-12 at 8 25 47 am

Although now that I think about it some more, this probably also includes all dependencies, such as inline-style-prefixer, which might all add up to somewhere near this size in the end.

Issue #239, opened 2017-05-10
Configure prefixer via browserslist


lencioni:
It looks like everybody is jumping on the browserslist train. I think it would be nice if we made it so the prefixes generated by Aphrodite could be configured in the same way that other tools like babel-preset-env and autoprefixer are configured.

More information: https://evilmartians.com/chronicles/autoprefixer-7-browserslist-2-released

BabakMN:
I'd also like to be able to disable autoprefixing during development. It makes debugging easier (in the console).

Kerumen:
I totally agree with @BabakMN. Currently, it's a pain to debug if you want to quickly disable / enable a CSS property as you have to do for all the prefixed-properties..

cezaraugusto:
is there an intention to implement for this?

PR #281, opened 2017-10-12
Build with rollup instead of webpack

lencioni:
Rollup will produce a smaller and more optimized bundle than webpack,
and can be configured in a way that works perfectly for libraries, such
as Aphrodite. This will help to minimize the bundle size impact of using
this package, and may even give a small runtime speedup. For reference,
React 16 is built using Rollup.

Rollup does not allow Babel 5, so I also updated to Babel 6 at the same
time. In this update, I tried to take care to maintain the same list of
browser support that we have listed in the CSS prefixes that we build.
At some point, we probably want to unify this configuration via
browserslist. Issue:

  https://github.com/Khan/aphrodite/issues/239

Along with this update, I decided to add an ES modules build since it
was easy enough. This will be used automatically by tools such as
webpack 2+ to import the ES6 module version directly. I think it still
makes sense to run these through Babel since most people don't run their
node_modules through Babel, so the main difference here is that the
import/export statements will not be compiled to require/module.exports.
This allows webpack to perform optimizations such as tree-shaking and
scope hoisting.

One risk to be on the lookout for when people update to this version is
that if you are using `require` to bring in Aphrodite with a version of
webpack that is ES modules capable, it will break. Those consumers will
need to switch to import instead. For this reason, I would be okay with
removing the `module` field from the package.json for an initial release
of the rollup build, and then we can add it later when the ecosystem has
time to catch up. This is the approach we landed on for react-waypoint:

1. https://github.com/brigade/react-waypoint/pull/220
2. https://github.com/brigade/react-waypoint/pull/223

The filesize of the dist builds before this change looked like:

```
 83K aphrodite.js
 84K aphrodite.umd.js
104K aphrodite.umd.js.map
 23K aphrodite.umd.min.js
204K aphrodite.umd.min.js.map
```

And after this change:

```
 72K aphrodite.js
 73K aphrodite.umd.js
108K aphrodite.umd.js.map
 20K aphrodite.umd.min.js
 93K aphrodite.umd.min.js.map
```

So it looks like the minified UMD build dropped from 24 KiB to 20 KiB.

cc @vladnicula


lencioni:
Looks like CI is failing because of dropped coverage for some reason. I'm guessing something about the babel update. I'll wait for feedback on this before I look into fixing that.

lencioni:
Ok, I think I fixed the coverage. Turns out I needed to bring in the istanbul babel plugin so that the coverage would be calculated correctly. PTAL @xymostech!

.babelrc
```diff
@@ -0,0 +1,22 @@
+{
```
xymostech: What babel configuration were we using before? Can we try to stay consistent with that?

.babelrc
```diff
@@ -0,0 +1,22 @@
...
+      "ie": "9",
+      "edge": "12",
+      "opera": "13",
+    },
+  }]],
+
+  "env": {
+    "test": {
+      "presets": ["airbnb"],
+      "plugins": ["istanbul"],
```
xymostech: What does this do?

tests/ordered-elements_test.js
```diff
@@ -2,95 +2,95 @@ import {assert} from 'chai';
 
 import OrderedElements from '../src/ordered-elements';
 
-describe("OrderedElements", () => {
-    it("can identify elements it has", () => {
+describe('OrderedElements', () => {
```
xymostech: not sure why all these quotes changed

rollup.config.js
```diff
@@ -0,0 +1,72 @@
...
+            commonjs(), // so rollup can convert node modules to ESM if needed
+        ]
+    };
+}
+
+export default [
+    distBuild({ filename: 'aphrodite.umd.js', format: 'umd', sourcemap: true, minify: false }),
+    distBuild({ filename: 'aphrodite.umd.min.js', format: 'umd', sourcemap: true, minify: true }),
+    distBuild({ filename: 'aphrodite.js', format: 'cjs', sourcemap: false, minify: false }),
+    standardBuilds('index'),
```
xymostech: just to check my understanding, is this going to build bundles into `lib/` or leave things in individual files like they are currently?

.babelrc
lencioni: Before this PR, Aphrodite was using a really old version of Babel, which was before presets existed and I believe Babel had a bunch of things turned on by default. When Babel 6 came out 2 years ago, everything changed and requires a lot more manual configuration.

I think the configuration I have here is going to be pretty consistent with what we had before, but I also made an effort to copy the browser targets we use for the CSS that is generated, which makes more sense to me than trying to stay consistent with the previous version of aphrodite. That way the JS and CSS is built to target the same sets of devices.

.babelrc
lencioni: This allows the code coverage to be properly calculated on the pre-compiled code. Without this, the code coverage checks failed. https://github.com/istanbuljs/babel-plugin-istanbul

rollup.config.js
```diff
@@ -0,0 +1,72 @@
...
+        ]
+    };
+}
+
+function standardBuilds(filename) {
+    return {
+        input: `src/${filename}.js`,
+        external: [
+            ...Object.keys(pkg.dependencies),
+            'inline-style-prefixer/static/createPrefixer',
```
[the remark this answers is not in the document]
lencioni: I'm not a rollup expert, but I wasn't able to find a way initially. I'll dig into this more.

rollup.config.js
```diff
@@ -0,0 +1,72 @@
...
+            'inline-style-prefixer/static/plugins/flexboxOld',
+            'inline-style-prefixer/static/plugins/gradient',
+            'inline-style-prefixer/static/plugins/imageSet',
+            'inline-style-prefixer/static/plugins/position',
+            'inline-style-prefixer/static/plugins/sizing',
+            'inline-style-prefixer/static/plugins/transition',
+        ],
+        output: [
+            { file: `lib/${filename}.js`, format: 'cjs' },
+            { file: `lib/${filename}.mjs`, format: 'es' },
```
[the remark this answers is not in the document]
lencioni: This will build a version that does not compile `import` to `require`, which enables build tools that can understand `import` (e.g. webpack 2+) to perform additional optimizations such as "tree-shaking" and scope hoisting.

```sh
~/aphrodite ❯❯❯ head lib/index.js
'use strict';

Object.defineProperty(exports, '__esModule', { value: true });

function _interopDefault (ex) { return (ex && (typeof ex === 'object') && 'default' in ex) ? ex['default'] : ex; }

var createPrefixer = _interopDefault(require('inline-style-prefixer/static/createPrefixer'));
var calc = _interopDefault(require('inline-style-prefixer/static/plugins/calc'));
var crossFade = _interopDefault(require('inline-style-prefixer/static/plugins/crossFade'));
var cursor = _interopDefault(require('inline-style-prefixer/static/plugins/cursor'));

~/aphrodite ❯❯❯ head lib/index.mjs
import createPrefixer from 'inline-style-prefixer/static/createPrefixer';
import calc from 'inline-style-prefixer/static/plugins/calc';
import crossFade from 'inline-style-prefixer/static/plugins/crossFade';
import cursor from 'inline-style-prefixer/static/plugins/cursor';
import filter from 'inline-style-prefixer/static/plugins/filter';
import flex from 'inline-style-prefixer/static/plugins/flex';
import flexboxIE from 'inline-style-prefixer/static/plugins/flexboxIE';
import flexboxOld from 'inline-style-prefixer/static/plugins/flexboxOld';
import gradient from 'inline-style-prefixer/static/plugins/gradient';
import imageSet from 'inline-style-prefixer/static/plugins/imageSet';
```

The .mjs extension is the proposed way to specify this type of build that Node will eventually adopt, but it is really early to see it in many places. I'd be just as happy to build this somewhere like `es/${filename}.js` if you prefer, and that might be a bit smoother for folks anyway until the dust settles on the .mjs extension.

rollup.config.js
lencioni: This will build bundles into `lib/` without the individual files. Here's what my `lib/` directory looks like after building like this:

```
total 504
drwxr-xr-x  12 joe_lencioni  staff   384B Oct 12 11:30 .
drwxr-xr-x  31 joe_lencioni  staff   992B Oct 13 10:54 ..
-rw-r--r--   1 joe_lencioni  staff   5.8K Oct 12 10:46 exports.js
-rw-r--r--   1 joe_lencioni  staff    14K Oct 12 10:46 generate.js
-rw-r--r--   1 joe_lencioni  staff    46K Oct 12 11:30 index.js
-rw-r--r--   1 joe_lencioni  staff    45K Oct 12 11:30 index.mjs
-rw-r--r--   1 joe_lencioni  staff    11K Oct 12 10:46 inject.js
-rw-r--r--   1 joe_lencioni  staff    45K Oct 12 11:30 no-important.js
-rw-r--r--   1 joe_lencioni  staff    44K Oct 12 11:30 no-important.mjs
-rw-r--r--   1 joe_lencioni  staff   4.4K Oct 12 10:46 ordered-elements.js
-rw-r--r--   1 joe_lencioni  staff   4.4K Oct 12 10:46 staticPrefixData.js
-rw-r--r--   1 joe_lencioni  staff   6.0K Oct 12 10:46 util.js
```

src/no-important.js
```diff
@@ -2,11 +2,26 @@
 // Module with the same interface as the core aphrodite module,
 // except that styles injected do not automatically have !important
 // appended to them.
-import {defaultSelectorHandlers} from './generate';
+import { defaultSelectorHandlers } from './generate';
```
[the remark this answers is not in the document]
lencioni: Ah, my bad.

tests/ordered-elements_test.js
lencioni: Just making things consistent. I can change this back if you like.

lencioni:
@xymostech can you take another look at this when you have a moment?

lencioni:
@xymostech since we decided that #240 is a breaking change, I'd like to ship this at the same time. Can you look at this again and let me know if you see anything you'd like changed about it?

rollup.config.js
lencioni: I've updated this PR to build the cjs version into lib/ and the esm version into es/

lencioni:
Going to merge now to keep things moving forward. Please let me know if you'd like me to follow up with any changes!

PR #291, opened 2017-12-08
Add minify global function to force style names minification

### LINKED ISSUE #288: Minify classnames as an option
It would be great to have a way to tell aphrodite not to minify classnames.
We currently have our classnames minified in production, because we set `process.env.NODE_ENV = 'production'`.
The said option would prevail on `process.env.NODE_ENV` value.

--- 

This option would prove useful for us to debug errors in production. We are using a tool that notifies about errors in our app. The tool adds an history of interaction with the different HTML elements as context. But we learn nothing from `User clicked on div._x23hsa`

screen shot 2017-11-10 at 15 11 58

Thank you for the great tool, by the way. It feels good to forget about the "cascading" part of CSS.

Soreine:
Fix #288 

This exports a global function `minify` that can be called to control the minification of style names, independently of the `NODE_ENV` environment value.

Call `minify(false)` before using `StyleSheet.create` to never minify style names. Call `minify(true)` to always minify style names. Defaults to the current behavior of minifying only if `process.env.NODE_ENV === 'production'`.

### DIFF
--- a/src/exports.js
+++ b/src/exports.js
@@ -17,13 +17,26 @@ type Extension = {
 export type MaybeSheetDefinition = SheetDefinition | false | null | void
 */
 
+// True to minify classnames.
+// False to not minify classnames.
+// Unset to minify only in 'production' environment
+let forceMinify = undefined;
+const shouldMinify = () => {
+    if (forceMinify !== undefined) {
+        return forceMinify;
+    } else {
+        return process.env.NODE_ENV === 'production';
+    }
+}
+
+
 const StyleSheet = {
     create(sheetDefinition /* : SheetDefinition */) {
         return mapObj(sheetDefinition, ([key, val]) => {
             const stringVal = JSON.stringify(val);
             return [key, {
                 _len: stringVal.length,
-                _name: process.env.NODE_ENV === 'production' ?
+                _name: shouldMinify() ?
                     hashString(stringVal) : `${key}_${hashString(stringVal)}`,
                 _definition: val
             }];
@@ -130,6 +143,10 @@ const makeExports = (
 
         StyleSheetServer,
         StyleSheetTestUtils,
+
+        minify(shouldMinify /* : boolean */) {
+            forceMinify = shouldMinify;
+        },
 
         css(...styleDefinitions /* : MaybeSheetDefinition[] */) {
             return injectAndGetClassName(
--- a/README.md
+++ b/README.md
@@ -1,8 +1,8 @@
 # Aphrodite: Inline Styles that work
 
-[![npm version](https://badge.fury.io/js/aphrodite.svg)](https://badge.fury.io/js/aphrodite) 
-[![Build Status](https://travis-ci.org/Khan/aphrodite.svg?branch=master)](https://travis-ci.org/Khan/aphrodite) 
-[![Coverage Status](https://coveralls.io/repos/github/Khan/aphrodite/badge.svg?branch=master)](https://coveralls.io/github/Khan/aphrodite?branch=master) 
+[![npm version](https://badge.fury.io/js/aphrodite.svg)](https://badge.fury.io/js/aphrodite)
+[![Build Status](https://travis-ci.org/Khan/aphrodite.svg?branch=master)](https://travis-ci.org/Khan/aphrodite)
+[![Coverage Status](https://coveralls.io/repos/github/Khan/aphrodite/badge.svg?branch=master)](https://coveralls.io/github/Khan/aphrodite?branch=master)
 [![Gitter chat](https://img.shields.io/gitter/room/Khan/aphrodite.svg)](https://gitter.im/Khan/aphrodite)
 
 [![gzip size][gzip-badge]][unpkg-dist]
@@ -183,6 +183,23 @@ to avoid this behaviour, then instead of importing `aphrodite`, import
 
 ```js
 import { StyleSheet, css } from 'aphrodite/no-important';
+```
+
+## Minifying style names
+
+By default, Aphrodite will minify style names down to their hashes in production
+(`process.env.NODE_ENV === 'production'`). You can override this behavior by
+calling `minify` with `true` or `false` before calling `StyleSheet.create`.
+
+This is useful if you want to facilitate debugging in production for example.
+
+```js
+import { StyleSheet, minify } from 'aphrodite';
+
+// Always keep the full style names
+minify(false);
+
+// ... proceed to use StyleSheet.create etc.
 ```
 
 ## Font Faces
--- a/tests/index_test.js
+++ b/tests/index_test.js
@@ -6,6 +6,7 @@ import {
   StyleSheet,
   StyleSheetServer,
   StyleSheetTestUtils,
+  minify,
   css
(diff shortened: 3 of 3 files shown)


Soreine:
`[clabot:check]`

src/exports.js
```diff
@@ -17,13 +17,26 @@ type Extension = {
...
+
+
 const StyleSheet = {
     create(sheetDefinition /* : SheetDefinition */) {
         return mapObj(sheetDefinition, ([key, val]) => {
             const stringVal = JSON.stringify(val);
             return [key, {
                 _len: stringVal.length,
-                _name: process.env.NODE_ENV === 'production' ?
+                _name: shouldMinify() ?
```
lencioni: With the previous version of this code, the conditional will end up being minified out in production, which gives us a small performance boost. Switching this to a runtime check will affect performance some amount.

It is probably a pretty small hit, but given that this code is in a pretty performance-sensitive path, it is definitely worth profiling the fully simplified version (e.g. `_name: hashString(stringVal)`) against this version in real-world scenarios so we at least understand what the performance impact of this change is.

Can you do this profiling and post your findings on this PR?

Soreine: That's interesting... Do you have ready to use profilers ?

lencioni: Not that I am aware of. I think to try this out, it would probably be best to use `npm link` in a project that is setting the `NODE_ENV` environment variable to production and minifying it out properly (e.g. in conjunction with [webpack's DefinePlugin](https://webpack.js.org/plugins/define-plugin/)).

Soreine:
Update:
I'm sorry, but I don't have the time to dive into performance benchmarks right now. If anyone knows webpack's DefinePlugin and would easily be able to do the benchmarks suggested by @lencioni I would be very grateful.

Maybe another approach would be to measure the number of times this code path is executed in a real-case scenario, then make a separate benchmark that measures the performance of the compiled equivalent of `process.env.NODE_ENV === 'production'` check versus the compiled runtime check. From these two infos, we can derive a measure of the performance in a real-case scenario.

dmiller9911:
I went ahead and ran some benchmarks for this.  I went ahead and used the same benchmarks that [react-native-web](https://github.com/necolas/react-native-web/tree/master/packages/benchmarks) has set up.  I made 1 change to the benchmark for Aphrodite. I moved `StyleSheet.create` call into the render function. This resulted in the create call to be called 189,400 times each run. This is a little exaggerated since most apps will have this outside the render function, resulting in a much lower amount of calls. Below are the results of the benchmarks:

All measurements in `ms`

#### Current
|      |DeepTree|        |        |WideTree |         |        | 
|------|--------|--------|--------|---------|---------|--------| 
| Run# | Median | Mean   | STDEV  | Median  | Mean    | STDEV  | 
| 1    | 83.86  | 91.24  | 25.45  | 138.88  | 137.96  | 14.68  | 
| 2    | 73.11  | 75.65  | 17.81  | 135.88  | 135.82  | 16.98  | 
| 3    | 75.5   | 78.28  | 21.3   | 124.82  | 126.94  | 8.99   | 
| 4    | 78.18  | 81.2   | 19.11  | 133.25  | 132.47  | 11.36  | 
| 5    | 78.02  | 80.22  | 23.27  | 137.2   | 134.68  | 12.15  | 
| Avg  | 77.734 | 81.318 | 21.388 | 134.006 | 133.574 | 12.832 | 

#### Fork No Changes
|      |DeepTree|        |        |WideTree |         |        | 
|------|--------|--------|--------|---------|---------|--------| 
| Run# | Median | Mean   | STDEV  | Median  | Mean    | STDEV  | 
| 1    | 73.26  | 78.89  | 18.55  | 131.28  | 129.32  | 11.08  | 
| 2    | 78.58  | 85.81  | 21.13  | 133.29  | 131.41  | 12.98  | 
| 3    | 79.85  | 83.97  | 19.84  | 127.98  | 128.75  | 8.65   | 
| 4    | 77.73  | 80.59  | 20.29  | 126.12  | 129.27  | 14.31  | 
| 5    | 81.2   | 84.5   | 19.72  | 121.97  | 127.1   | 10.17  | 
| Avg  | 78.124 | 82.752 | 19.906 | 128.128 | 129.17  | 11.438 | 


#### Fork minify(true)
|      |DeepTree|        |        |WideTree |         |        | 
|------|--------|--------|--------|---------|---------|--------|  
| Run# | Median | Mean   | STDEV  | Median  | Mean    | STDEV  | 
| 1    | 78.01  | 82.11  | 20.3   | 132.83  | 134.27  | 16.31  | 
| 2    | 76.93  | 79.11  | 20.27  | 128.2   | 129.77  | 12.47  | 
| 3    | 75.53  | 78.61  | 18.68  | 131.48  | 132.1   | 11.27  | 
| 4    | 77.79  | 83.44  | 24.24  | 133.77  | 133.86  | 11.53  | 
| 5    | 63.09  | 69.78  | 19.57  | 125.24  | 126.04  | 7      | 
| Avg  | 74.27  | 78.61  | 20.612 | 130.304 | 131.208 | 11.716 | 

 
#### Fork minify(false)
|      |DeepTree|        |        |WideTree |         |        | 
|------|--------|--------|--------|---------|---------|--------| 
| Run# | Median | Mean   | STDEV  | Median  | Mean    | STDEV  | 
| 1    | 88.32  | 90.17  | 23.77  | 124.22  | 132.03  | 11.17  | 
| 2    | 79.62  | 85     | 22.64  | 133.74  | 130.26  | 9.82   | 
| 3    | 74.24  | 81.43  | 22.13  | 131.51  | 137.91  | 14.98  | 
| 4    | 79.04  | 82.64  | 20.76  | 128.43  | 132.51  | 15.62  | 
| 5    | 80.3   | 84.12  | 20.23  | 135.13  | 135.03  | 12.08  | 
| Avg  | 80.304 | 84.672 | 21.906 | 130.606 | 133.548 | 12.734 | 

There is only a slight difference between each scenario. My guess is if I ran the benchmark more times for each they would end up much closer results. I do not think that having 1 more if statement will really impact performance much, especially inside the `StyleSheet.create` call.

lencioni:
@Soreine this seems to fail tests in CI after merging. https://travis-ci.org/Khan/aphrodite/builds/341685191

Any chance you can take a look and open a PR with a fix?

Soreine:
@lencioni yes, sir.

Kerumen:
@lencioni Any chance to have this released as an npm tag? Or this library is not maintained anymore?

Soreine:
@Kerumen this library is still maintained :) I think they just forgot, were busy or were waiting on something

Kerumen:
@Soreine The latest release was last October (https://github.com/Khan/aphrodite/releases). We are now in February and 30 commits are awaiting a release...

dmiller9911:
Another release is coming soon I believe. I think the maintainers are just waiting for #300 to go in and then a major release will be created.

lencioni:
@Kerumen I hope to publish a release soon. This library is maintained and the release has not been forgotten. The next release will be a major version bump, so I want to make sure we get in all of the in-flight breaking changes before we publish.

PR #297, opened 2018-02-06
Bump inline-style-prefixer to 4.0.0

calinoracation:
This bumps the version of [inline-style-prefixer](https://github.com/rofrischmann/inline-style-prefixer) to pick up the bug fixes and performance improvements introduced since 3.0.1, which includes a reported 30% improvement. From what I can tell, the major version bump was due to them changing their default browser prefix list, which we specify our own in https://github.com/Khan/aphrodite/blob/master/tools/generate_prefixer_data.js so I don't believe that will be an issue. https://github.com/rofrischmann/inline-style-prefixer/blob/master/Changelog.md

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "aphrodite",
-  "version": "1.2.1",
+  "version": "1.2.5",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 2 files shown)


lencioni:
@rofrischmann how much of the performance improvement do you think comes from updating the default browser list vs other types of changes in v4?

I wonder if it is worth allowing either v3 or v4 in the dependency range?

calinoracation:
It looks like the change for the 30% perf improvement was introduced in 3.0.2, but a potential bug in the ordering wasn't fixed until 4.0.0 in https://github.com/rofrischmann/inline-style-prefixer/issues/147.

lencioni:
Nice, thanks for the numbers!

calinoracation:
I'll add more extensive ones that use puppeteer and can be run directly from this repo in a follow-up PR, just wanted to post a few initial numbers!

lencioni:
Awesome!

PR #298, opened 2018-02-15
Fix export of minify global function

Soreine:
Fix error following merge of the `minify` PR. The way `index.js` exports were made has been changed on `master` since I had started the PR.
See https://github.com/Khan/aphrodite/pull/291#issuecomment-365788953


lencioni:
Thanks!

PR #300, opened 2018-02-16
Replace object spread a reduce in generateCSSRuleset

lencioni:
I was profiling Aphrodite and noticed that the self time in
generateCSSRuleset is one of the most expensive bits. Digging in a bit,
it looked like most of this was coming from the object spread here,
which we use do copy the original object so we can reference its keys
with fast lookups. By swapping this out for a ~Set~ reduce we greatly reduce the
self time of this function from ~7% of total runtime to ~3%.

~I've implemented this in a way that expects Set to be available in the
browser or polyfilled, which I think is a reasonable explanation. I've
documented this in the readme. If we move forward with this, we could
also make Map a requirement and simplify a couple of codepaths if we
want.~

### DIFF
--- a/src/generate.js
+++ b/src/generate.js
@@ -250,6 +250,11 @@ const transformRule = (
 );
 
 
+const arrayToObjectKeysReducer = (acc, val) => {
+    acc[val] = true;
+    return acc;
+};
+
 /**
  * Generate a CSS ruleset with the selector and containing the declarations.
  *
@@ -291,7 +296,8 @@ export const generateCSSRuleset = (
     // Mutates declarations
     runStringHandlers(declarations, stringHandlers, selectorHandlers);
 
-    const originalElements = new Set(Object.keys(declarations.elements));
+    const originalElements = Object.keys(declarations.elements)
+        .reduce(arrayToObjectKeysReducer, Object.create(null));
 
     // NOTE(emily): This mutates handledDeclarations.elements.
     const prefixedElements = prefixAll(declarations.elements);
@@ -305,7 +311,7 @@ export const generateCSSRuleset = (
         // sortOrder, which means it was added by prefixAll. This means that we
         // need to figure out where it should appear in the sortOrder.
         for (let i = 0; i < elementNames.length; i++) {
-            if (!originalElements.has(elementNames[i])) {
+            if (!originalElements[elementNames[i]]) {
                 // This element is not in the sortOrder, which means it is a prefixed
                 // value that was added by prefixAll. Let's try to figure out where it
                 // goes.
@@ -324,7 +330,7 @@ export const generateCSSRuleset = (
                     originalStyle = elementNames[i][2].toLowerCase() + elementNames[i].slice(3);
                 }
 
-                if (originalStyle && originalElements.has(originalStyle)) {
+                if (originalStyle && originalElements[originalStyle]) {
                     const originalIndex = declarations.keyOrder.indexOf(originalStyle);
                     declarations.keyOrder.splice(originalIndex, 0, elementNames[i]);
                 } else {
--- a/src/generate.js
+++ b/src/generate.js
@@ -291,7 +291,7 @@ export const generateCSSRuleset = (
     // Mutates declarations
     runStringHandlers(declarations, stringHandlers, selectorHandlers);
 
-    const originalElements = {...declarations.elements};
+    const originalElements = new Set(Object.keys(declarations.elements));
 
     // NOTE(emily): This mutates handledDeclarations.elements.
     const prefixedElements = prefixAll(declarations.elements);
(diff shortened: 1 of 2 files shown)


lencioni:
@xymostech I think this would be worth including in the next release 💃

lencioni:
I ran this through https://github.com/hellofresh/css-in-js-perf-tests and here are my results.

## master

```
Running simple test.

  aphrodite              x 30,105 ops/sec ±1.28% (81 runs sampled)
  aphrodite-no-important x 31,924 ops/sec ±0.94% (86 runs sampled)

Running nested test.

  aphrodite              x 12,276 ops/sec ±2.35% (85 runs sampled)
  aphrodite-no-important x 13,814 ops/sec ±1.48% (87 runs sampled)

Running style overload test.

  aphrodite              x 3,991 ops/sec ±4.13% (80 runs sampled)
  aphrodite-no-important x 4,059 ops/sec ±12.48% (77 runs sampled)

Running classes overload test.

  aphrodite              x 4,526 ops/sec ±3.02% (78 runs sampled)
  aphrodite-no-important x 5,095 ops/sec ±4.26% (79 runs sampled)

```

## This branch

```
Running simple test.

  aphrodite              x 29,719 ops/sec ±1.22% (80 runs sampled)
  aphrodite-no-important x 32,488 ops/sec ±1.16% (87 runs sampled)


Running nested test.

  aphrodite              x 12,535 ops/sec ±1.55% (82 runs sampled)
  aphrodite-no-important x 14,301 ops/sec ±0.93% (89 runs sampled)


Running style overload test.

  aphrodite              x 4,153 ops/sec ±1.80% (81 runs sampled)
  aphrodite-no-important x 4,631 ops/sec ±1.62% (86 runs sampled)

Running classes overload test.

  aphrodite              x 5,242 ops/sec ±1.28% (83 runs sampled)
  aphrodite-no-important x 5,534 ops/sec ±1.76% (84 runs sampled)

```

src/generate.js
```diff
@@ -291,7 +291,7 @@ export const generateCSSRuleset = (
     // Mutates declarations
     runStringHandlers(declarations, stringHandlers, selectorHandlers);
 
-    const originalElements = {...declarations.elements};
+    const originalElements = new Set(Object.keys(declarations.elements));
```
dmiller9911: Might be nice to add a check for this in dev since browser support is advertised at IE9 [/tools/generate_prefixer_data.js#L14](https://github.com/Khan/aphrodite/blob/master/tools/generate_prefixer_data.js#L14). This could be missed be some and not caught until consumers have reported it.

 Another possibility is to make this a major release since it is technically breaking for IE9 and 10.

lencioni: The next release will be a major version anyway, so we should be alright here. I like the idea of a check in dev but not really sure I see how that would be much more useful than it just erroring out normally. Thoughts?

dmiller9911: Major will be perfect.  Definitely looking forward to the next release. 

For the dev check, I was thinking more of a console warning.  Since I am sure most do not do local dev with <IE11, although I am not sure that it would add any benefit.  I was more hoping for the major version increase.

lencioni:
I think I have a different method using `Object.create(null)` and a reducer that has similar (and potentially better) perf characteristics than `Set`, but without being a breaking change. Here's the results from the css-in-js-perf-tests.

## 1.2.5

```
Running simple test.

  1 test completed.

  aphrodite x 24,003 ops/sec ±1.19% (82 runs sampled)


Running nested test.

  1 test completed.

  aphrodite x 10,371 ops/sec ±2.66% (78 runs sampled)


Running style overload test.

  1 test completed.

  aphrodite x 3,461 ops/sec ±1.31% (83 runs sampled)


Running classes overload test.

  1 test completed.

  aphrodite x 4,069 ops/sec ±1.85% (79 runs sampled)
```

## master

```
Running simple test.

  1 test completed.

  aphrodite x 23,930 ops/sec ±1.94% (84 runs sampled)


Running nested test.

  1 test completed.

  aphrodite x 10,265 ops/sec ±2.25% (81 runs sampled)


Running style overload test.

  1 test completed.

  aphrodite x 3,433 ops/sec ±2.43% (80 runs sampled)


Running classes overload test.

  1 test completed.

  aphrodite x 3,976 ops/sec ±2.68% (81 runs sampled)
```

## Set

```
Running simple test.

  1 test completed.

  aphrodite x 25,067 ops/sec ±1.76% (83 runs sampled)


Running nested test.

  1 test completed.

  aphrodite x 10,425 ops/sec ±2.05% (82 runs sampled)


Running style overload test.

  1 test completed.

  aphrodite x 3,513 ops/sec ±2.00% (79 runs sampled)


Running classes overload test.

  1 test completed.

  aphrodite x 4,311 ops/sec ±2.27% (83 runs sampled)
```

## Object.create

```
Running simple test.

  1 test completed.

  aphrodite x 24,126 ops/sec ±1.88% (83 runs sampled)


Running nested test.

  1 test completed.

  aphrodite x 10,331 ops/sec ±1.73% (83 runs sampled)


Running style overload test.

  1 test completed.

  aphrodite x 3,397 ops/sec ±2.02% (81 runs sampled)


Running classes overload test.

  1 test completed.

  aphrodite x 4,057 ops/sec ±2.27% (82 runs sampled)
```

Profiling in Chrome with a pretty naive benchmark script, it looks like the `Object.create` approach is marginally faster than `Set`. I'll push up a commit soon with this change.

PR #301, opened 2018-02-21
Windows Development Support

dmiller9911:
Use cross-env and add node to build:prefixes to support developers on windows.


package.json
```diff
@@ -65,6 +65,7 @@
   },
   "dependencies": {
     "asap": "^2.0.3",
+    "cross-env": "^5.1.3",
```
lencioni: This should be a devDependency I believe

lencioni:
Thanks!

PR #302, opened 2018-02-21
Moved typings into repo, and updated typings to improve type safety.

dmiller9911:
Replaces #258 since the contributor has asked for someone else to take over.

* Provides intellisense for propteries in StyleSheet.Create
* Provides type safety for fonts
* updated to include new minify function and StyleSheetTestUtils.getBufferedStyles
* Removed dependency on React.CSSProperties typings by pulling in typings from glamorous, and editing to match Aphrodite.

### DIFF
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -1,20 +1,16 @@
 import {
     CSSProperties,
-    CSSPropertiesArray,
     CSSPropertiesComplete,
     CSSPropertiesLossy,
     CSSPropertiesPseudo,
-    CSSPropertiesRecursive,
     CSSWideKeyword,
 } from './css-properties';
 
 export {
     CSSProperties,
-    CSSPropertiesArray,
     CSSPropertiesComplete,
     CSSPropertiesLossy,
     CSSPropertiesPseudo,
-    CSSPropertiesRecursive,
     CSSWideKeyword,
 };
 
@@ -23,19 +19,13 @@ export {
  */
 export type StyleDeclarationMap = Map<keyof CSSProperties, string | number>;
 export type StyleDeclaration<T = {}> = {
-    [P in keyof T]: CSSProperties | StyleDeclarationMap
+    [P in keyof T]: CSSProperties | StyleDeclarationMap;
 };
 
 /**
  * Return value from StyleSheet.create.
  */
-export interface StyleDeclarationValue<
-    T extends CSSProperties | StyleDeclarationMap = {}
-> {
-    _len: number;
-    _name: string;
-    _definition: T;
-}
+export type StyleDeclarationValue = object;
 
 interface StyleSheetStatic {
     /**
@@ -43,7 +33,7 @@ interface StyleSheetStatic {
      */
     create<T extends StyleDeclaration<T>>(
         styles: T
-    ): { [K in keyof T]: StyleDeclarationValue<T[K]> };
+    ): {[K in keyof T]: StyleDeclarationValue };
     /**
      * Rehydrate class names from server renderer
      */
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -1,0 +1,133 @@
+import {
+    CSSProperties,
+    CSSPropertiesArray,
+    CSSPropertiesComplete,
+    CSSPropertiesLossy,
+    CSSPropertiesPseudo,
+    CSSPropertiesRecursive,
+    CSSWideKeyword,
+} from './css-properties';
+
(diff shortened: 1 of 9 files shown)


jlfwong:
This is awesome ✨!

I'll give this a try tonight by using it a TypeScript project I have. Thanks for taking this on :)

typings/index.d.ts
```diff
@@ -0,0 +1,133 @@
...
+    ): { [K in keyof T]: StyleDeclarationValue<T[K]> };
+    /**
+     * Rehydrate class names from server renderer
+     */
+    rehydrate(renderedClassNames: string[]): void;
+}
+
+export var StyleSheet: StyleSheetStatic;
+
+type CSSInputTypes = StyleDeclarationValue | false | null | void;
```
jlfwong: I think you can either do ` | false | void` or do ` | false | null | undefined`. I believe `void` is equivalent to `null | undefined`.

typings/index.d.ts
```diff
@@ -0,0 +1,133 @@
...
+    CSSPropertiesRecursive,
+    CSSWideKeyword,
+};
+
+/**
+ * Aphrodite style declaration
+ */
+export type StyleDeclarationMap = Map<keyof CSSProperties, string | number>;
+export type StyleDeclaration<T = {}> = {
+    [P in keyof T]: CSSProperties | StyleDeclarationMap
```
jlfwong: I didn't realize that aphrodite now explicitly supports calls like:

```
const declMap = new Map()
declMap.set("background", "red");
StyleSheet.create({
  foo: declMap
})
```

Does that seem right to you, @lencioni?

typings/index.d.ts
```diff
@@ -0,0 +1,133 @@
...
+};
+
+/**
+ * Return value from StyleSheet.create.
+ */
+export interface StyleDeclarationValue<
+    T extends CSSProperties | StyleDeclarationMap = {}
+> {
+    _len: number;
```
jlfwong: Since this is part of the private API, I don't know if it's valuable to export these properties, since this is intentionally not part of the public API.

typings/index.d.ts
```diff
@@ -0,0 +1,133 @@
...
+     *
+     * @returns {string}  Buffer of styles which have not yet been flushed.
+     */
+    getBufferedStyles(): string[];
+}
+
+export var StyleSheetTestUtils: StyleSheetTestUtilsStatic;
+
+export interface SelectorHandler {
+    (selector: string, baseSelector: string, callback: (selector: string) => string):
```
jlfwong: Am I misread this like? This looks like a function with a return type of `| string | null` (with the leading `|`) Is this supposed to be just `string | null`?

jlfwong:
I'm also unsure whether it's kosher to just link to the GitHub page for glamorous, or if we need to include the full MIT license in that file.

dmiller9911:
Yeah, I was unsure about the link to glamorous.  I followed their lead with it.  Their typing file has a link to the react typings since they use that as a base.  I also made some changes to the file to better align with Aphrodite. @lencioni any idea on the best way to handle that?

typings/index.d.ts
lencioni: Yep, this was added mostly to allow developers to ensure the order of styles in older versions of node. We needed it at Airbnb when it was introduced, but since then we have been able to update our node version to a new enough version that preserves key order in objects so we no longer need this feature. In practice, I'm not sure how many people are using it anymore.

lencioni:
I don't think it would hurt to include the license in the file, but I have no idea if it is necessary. I bet @ljharb would know though.

lencioni:
@dmiller9911 is it possible to have something run in CI that ensures that these types (or as many as possible) are always accurate?

typings/index.d.ts
jlfwong: That sounds good to me :)

jlfwong:
@dmiller9911 I think will be good to merge after the update to remove references to private fields!

dmiller9911:
@lencioni that is a good idea.  Should not be too difficult.  I will get something up today or tomorrow. Along with the changes suggested by @jlfwong

typings/index.d.ts
dmiller9911: I actually had to put null back in.  While doing tests null was not explicitly allowed, and failed the `css(null)` test

jlfwong:
This is great @dmiller9911! Thanks for putting this together and responding gracefully to feedback :)

PR #304, opened 2018-03-03
Remove separate iteration for summing length of style defs

lencioni:
I've been profiling and noticed that injectAndGetClassName is a very hot
path so I've been thinking about ways to make it more efficient. I
noticed that we were iterating over things twice unnecessarily, so I've
moved a reduce function inline with our original iteration. This
improves my naive benchmark from ~2.94s to ~2.77s.

While I was at it, I removed an unnecessary object creation here, to hlp
improve efficiency and reduce the memory footprint.

I think it might be possible to further improve perf here by converting
the recursive function into an interative one, which I'm going to dig
into next.

cc @gilbox


lencioni:
Update: the iterative version seems to be quite a bit slower (I think due to the addition of array splicing).

Issue #296, opened 2018-02-01
Multiple @font-face with the same font family name


milesj:
How to support this correctly? When using this approach (using the same font family name is valid):

```
fontFamily: [
 {
      fontFamily: 'Circular',
      fontStyle: 'normal',
      fontWeight: 'normal',
      src: '../font/Circular_Air-Book.woff2',
    },
    {
      fontFamily: 'Circular',
      fontStyle: 'italic',
      fontWeight: 'normal',
      src: '../font/Circular_Air-Book_Italic.woff2',
    },
    {
      fontFamily: 'Circular',
      fontStyle: 'normal',
      fontWeight: '700',
      src: '../font/Circular_Air-Bold.woff2',
    },
    {
      fontFamily: 'Circular',
      fontStyle: 'normal',
      fontWeight: '300',
      src: '../font/Circular_Air-Light.woff2',
    },
    // ... others
],
```

It generates CSS like the following:

```css
font-family: "Circular","Circular","Circular","Circular",-apple-system,BlinkMacSystemFont,Helvetica Neue,sans-serif !important;
```

Which in turn, causes the same font to be loaded over and over, causing annoying consecutive white screen flashes.

<img width="1274" alt="screen shot 2018-02-01 at 1 52 58 pm" src="https://user-images.githubusercontent.com/143744/35705437-46819118-0757-11e8-88ea-b5e8ce8a4181.png">

lencioni:
I'm not very familiar with this portion of Aphrodite, but this looks like a bug to me (and if it's not a bug, it should be clarified via documentation). Would you be interested in digging a little more into this problem, and perhaps opening a PR to address it (either updating docs, or modifying code)?

milesj:
Don't have a solution yet, but you can see the compiled output here: https://github.com/milesj/aesthetic/blob/master/packages/aesthetic-adapter-aphrodite/tests/__snapshots__/NativeAdapter.test.js.snap#L17

And the example syntax: https://github.com/milesj/aesthetic/blob/master/tests/mocks.js#L20 AND https://github.com/milesj/aesthetic/blob/master/packages/aesthetic-adapter-aphrodite/tests/NativeAdapter.test.js#L123

PR #327, opened 2018-05-23
Fix multiple @font-face of the same name

milesj:
Fix for https://github.com/Khan/aphrodite/issues/296. Verified the fix locally in one of my projects by editing the node_modules file.

Sidenote, I'd highly suggest using Jest snapshots to verify buffered output instead of `assertStylesInclude`, which is extremely flakey.


milesj:
@lencioni Fixed formatting/spacing.

PR #330, opened 2018-05-28
Update typings to resolve #329

### LINKED ISSUE #329: TS4023 when trying to extend using typescript
Given the code
```javascript
import { StyleSheet as Aphrodite } from 'aphrodite';

const { StyleSheet, css } = Aphrodite.extend([{ 
  selectorHandler: (selector, baseSelector, generateSubtreeStyles) => {
   /* Extension stuff here */
    return null;
  }
}]);
export { StyleSheet, css}
```
Typescript throws the following error:
`[ts] Exported variable 'StyleSheet' has or is using name 'StyleSheetStatic' from external module "[root]/node_modules/aphrodite/typings/index" but cannot be named.`

The import line needs to contain StyleSheetStatic
```javascript
import { StyleSheet as Aphrodite, StyleSheetStatic } from 'aphrodite';
```
However this isn't exported in the typing file.
https://github.com/Khan/aphrodite/blob/b9f5ae2dcc7fc0abeb426ea6f7939978e36b90e4/typings/index.d.ts#L30
Also seems there is duplication in to add in extend method to the interface:
https://github.com/Khan/aphrodite/blob/b9f5ae2dcc7fc0abeb426ea6f7939978e36b90e4/typings/index.d.ts#L121-L123
Not sure that is intended

Sibz:
exported StyleSheetStatic

Moved `extend` method definition into first `StyleSheetStatic` interface definition + removed second `StyleSheetStatic` definition

Fixes #329

### DIFF
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -27,7 +27,7 @@ export type StyleDeclaration<T = {}> = {
  */
 export type StyleDeclarationValue = object;
 
-interface StyleSheetStatic {
+export interface StyleSheetStatic {
     /**
      * Create style sheet
      */
@@ -38,6 +38,8 @@ interface StyleSheetStatic {
      * Rehydrate class names from server renderer
      */
     rehydrate(renderedClassNames: string[]): void;
+    
+    extend(extensions: Extension[]): Exports;
 }
 
 export var StyleSheet: StyleSheetStatic;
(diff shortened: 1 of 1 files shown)


Sibz:
[clabot:check]

lencioni:
@dmiller9911 Can you please review this when you have a chance?

PR #343, opened 2018-10-01
--save no longer needed

JoeCortopassi:
`--save` is on by default as of [npm 5](https://blog.npmjs.org/post/161081169345/v500), so `npm install aphrodite` is functionally equivalent to `npm install --save aphrodite` now


JoeCortopassi:
[clabot:check]

kevinbarabash:
I didn't realize this.  `npm 5` was released in May 2017 which is a while.  I guess if people have issues we can always add a note about earlier npm versions.

Issue #306, opened 2018-03-10
Add public function to determine if an object is an aphrodite style or not


kevinbarabash:
I'm working on some components that have a `style` prop that's similar to the `style` prop in React Native components which can take either inline styles or styles from StyleSheet objects.  The components need to know how to disambiguate between the types of objects being passed in so that they can apply the slides correctly in the render method.

Proposed API:
```
function isAphroditeStyle(style: any) => boolean
```

lencioni:
We do this in react-with-styles-interface-aphrodite. https://github.com/airbnb/react-with-styles-interface-aphrodite/blob/e9c088d8998bed347ff5c26f54a22d8dfd633cb6/src/utils/separateStyles.js

Can you make a case for adding it to this project vs a separate project or something?

kevinbarabash:
If we change the internal implementation of Aphrodite it will break that code.  It seems like a generally useful function to have to support tooling.  We could do it in a separate package but that package should be part of this repo and have tests so in order to avoid breakages.

lencioni:
Sure, might as well I guess. Want to submit a PR? It would be great if your PR refactored any internal code that might benefit to also use this function.

If I understand tree-shaking correctly, we should be able to expose this as a named export and it won't affect consumer bundle sizes if they don't import it.

kevinbarabash:
If the function is use internally people will still get the code.

PR #349, opened 2018-11-12
Clearer error message

### LINKED ISSUE #265: Using an object instead of `StyleSheet.create` causes `css` to throw difficult to trace error
This was my first experience using Aphrodite:

I tried switching from Radium to Aphrodite. In Radium, styles are expressed as follows:

```
const styles = {
  container: {
    fontSize: 'large',
  },
  // ... etc ...
};
```

I didn't notice at first that Aphrodite requires the use of `StyleSheet.create({})` in the definition of styles and rather than changing the styles, I started switching all of the `style={styles.xyz}` jsx attributes to `className={css(styles.xyz)}` attributes. What resulted was the following difficult to trace error message:

```
TypeError: Cannot convert undefined or null to object
    at keys (<anonymous>)
    at Function.keys (preview.bundle.js:19097)
    at OrderedElements.addStyleType (preview.bundle.js:68070)
    at generateCSS (preview.bundle.js:67780)
    at injectStyleOnce (preview.bundle.js:68475)
    at injectAndGetClassName (preview.bundle.js:68570)
    at css (preview.bundle.js:68249)
    at SearchBar (preview.bundle.js:49733)
    at preview.bundle.js:28187
    at measureLifeCyclePerf (preview.bundle.js:27957)
```

I realize this is a type error (and we have tools for this!) but it was an unpleasant first introduction to an otherwise fine looking library.

galencorey:
closes #265 

Add a more explicit error message when someone tries to pass a plain object instead of a stylesheet into the `css()` method. 

Note: My editor went crazy with the double quotes. Let me know if you would like me to switch them back!

### DIFF
--- a/src/inject.js
+++ b/src/inject.js
@@ -243,6 +243,9 @@ export const addRenderedClassNames = (classNames /* : string[] */) => {
     });
 };
 
+const isValidStyleDefinition = (def /* : Object */) =>
+  "_definition" in def && "_name" in def && "_len" in def;
+
 const processStyleDefinitions = (
     styleDefinitions /* : any[] */,
     classNameBits /* : string[] */,
@@ -261,10 +264,12 @@ const processStyleDefinitions = (
                     definitionBits,
                     length,
                 );
-            } else {
+            } else if (isValidStyleDefinition(styleDefinitions[i])) {
                 classNameBits.push(styleDefinitions[i]._name);
                 definitionBits.push(styleDefinitions[i]._definition);
                 length += styleDefinitions[i]._len;
+            } else {
+                throw new Error("Invalid Style Definition: Styles should be defined using the StyleSheet.create method.")
             }
         }
     }
--- a/tests/index_test.js
+++ b/tests/index_test.js
@@ -143,6 +143,10 @@ describe('css', () => {
             done();
         });
     });
+
+    it("throws a useful error for invalid arguments", () => {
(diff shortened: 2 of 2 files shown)


galencorey:
[clabot:check]

src/inject.js
```diff
@@ -243,6 +243,9 @@ export const addRenderedClassNames = (classNames /* : string[] */) => {
     });
 };
 
+const isValidStyleDefinition = (def /* : Object */) =>
+  "_definition" in def && "_name" in def && "_len" in def;
```
kevinbarabash: I was hoping to make something like this public for #306.

src/inject.js
```diff
@@ -261,10 +264,12 @@ const processStyleDefinitions = (
                     definitionBits,
                     length,
                 );
-            } else {
+            } else if (isValidStyleDefinition(styleDefinitions[i])) {
                 classNameBits.push(styleDefinitions[i]._name);
                 definitionBits.push(styleDefinitions[i]._definition);
                 length += styleDefinitions[i]._len;
+            } else {
+                throw new Error("Invalid Style Definition: Styles should be defined using the StyleSheet.create method.")
```
kevinbarabash: Nice!

galencorey:
@kevinbarabash thank you for your review! reverted all the linter changes. Prettier has no chill sometimes.

Issue #335, opened 2018-06-29
Multiple aphrodite instances on page break styles


Mati365:
After linking two components with _Aphrodite_ instance themes called using `css([style1, style2])` are removed in `style[data-aphrodite]` tag.

PR #356, opened 2019-01-08
Add ability to suffix data-aphrodite to prevent collisions between multiple Aphrodite instances

rtibbles:
Provides a more systematic solution to #335


rtibbles:
[clabot:check]

rtibbles:
Guessing aphrodite is unmaintained at this point - closing this PR.

PR #357, opened 2019-01-28
Update inject.js

istvanmakary:
Based on the HTML 5 definitions, the type attribute is not required on the style tags any more.

https://www.w3schools.com/tags/att_style_type.asp

`In HTML5, the type attribute is no longer required for CSS. The default value is "text/css".`

Best,
Istvan


kevinbarabash:
I worry that this might break some older browsers that might not be fully compliant.

lencioni:
Thanks for the PR! But agreed, I'm not sure this is a good trade-off to make.

PR #358, opened 2019-02-08
Bump up inline-style-prefixer version

chenyejia:
This PR bumps the version of `inline-style-prefixer` up to 5.0.4 so that Aphrodite has the css logical plugin.

The imports in `inline-style-prefixer` changed from versions 4 to 5, so this PR also updates some of the import paths, and removes the `compatibility: true` option in `generateData`, which determined whether the plugin data generated used ES5 or ES2015 imports.

### DIFF
--- a/package.json
+++ b/package.json
@@ -72,7 +72,7 @@
   },
   "dependencies": {
     "asap": "^2.0.3",
-    "inline-style-prefixer": "^4.0.2",
+    "inline-style-prefixer": "^5.0.4",
     "string-hash": "^1.1.3"
   },
   "tonicExampleFilename": "examples/runkit.js",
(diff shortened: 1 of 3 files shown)


chenyejia:
[clabot:check]

somewhatabstract:
Does the loss of `compatibility` setting have any implications for Aphrodite usage after this lands?

lencioni:
@somewhatabstract it looks like the compatibility flag only swaps between require or import for inline-style-prefixer plugins. More info in this PR: https://github.com/Khan/aphrodite/pull/361#issue-254044665

https://github.com/rofrischmann/inline-style-prefixer/blob/608ac33bd65d6732915745e74521b6475e1916d8/modules/generator/index.js#L5-L10

Since the require statement it generates does not use `.default`, the compatibility option breaks the aphrodite build. 

TL;DR: I don't think this has any implications for Aphrodite usage after this lands.

lencioni:
@chenyejia thanks for the PR!

PR #363, opened 2019-02-19
Use deep import for inline-style-prefixer

lencioni:
After updating to 2.3.0, I noticed a large bundle size increase. I
believe this is due to switching this from a deep import to a named
import, so we were duplicating a bunch of code (the named import used
the es build, which brought in the es plugins, but our generated data
brought in the lib plugins).

As a quick fix, I think we can bring bundle sizes back down by using a
deep import here again.


lencioni:
Bundlephobia is much happier with this. https://bundlephobia.com/result?p=aphrodite@2.3.1

image

PR #364, opened 2019-02-19
Roll up inline-style-prefixer

lencioni:
Since v5 of inline-style-prefixer removed the dynamic generator, it only
works with static generator data. It seems unlikely that other
dependencies in an application will bring in inline-style-prefixer, so
we can likely reduce bundle size a little by rolling it up here.

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -34,6 +34,7 @@ function distBuild(options) {
 }
 
 const externals = new Set(Object.keys(pkg.dependencies));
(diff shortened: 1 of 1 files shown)


lencioni:
I think we can get more bundle size improvements by aliasing inline-style-prefixer/lib to inline-style-prefixer/es, but I was unable to get that to work with rollup, I think due to this bug: https://github.com/rollup/rollup-plugin-alias/issues/27

PR #373, opened 2019-04-06
Update inline-style-prefixer 5.0.4 → 5.1.0

lencioni:
Changes:

- added a plugin that transforms grid properties


lencioni:
It's not full grid support, but it should be helpful

PR #375, opened 2019-04-22
Make parameter typings for StyleSheet.create more typed

BenjaVR:
This is a small change in the typings, to make the `StyleSheet.create(...)` parameters more typed, and offer better intellisense/autocompletion.

**before:**

screencast_before

**after:**

screencast_after

---

The only thing I couldn't figure out is how to get the intellisense for pseudo selectors inside a mediaquery. Although you can use it without any errors, it's just the intellisense that is not complete. Anyone any ideas?

### DIFF
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -31,8 +31,8 @@ export interface StyleSheetStatic {
     /**
      * Create style sheet
      */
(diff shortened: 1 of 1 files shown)


BenjaVR:
[clabot:check]

PR #380, opened 2019-07-02
Allow already injected to be reset

milesj:
to: @lencioni 

We have some logic in our app that deletes injected CSS rules (mostly global styles). This works great.

However, once the styles are regenerate and attempt to inject again, they will not as Aphrodite blocks the injection because the previous key is set to `alreadyInjected`. https://github.com/Khan/aphrodite/blob/master/src/inject.js#L159

This PR adds a minor function called `resetInjected` to clear this cache. I also went ahead and exported `reset` since it may also be useful.

### DIFF
--- a/src/exports.js
+++ b/src/exports.js
@@ -3,6 +3,7 @@ import {hashString} from './util';
 import {
     injectAndGetClassName,
     reset,
+    resetInjected,
     startBuffering,
     flushToString,
     flushToStyleTag,
@@ -188,5 +189,7 @@ export default function makeExports(
         flushToStyleTag,
         injectAndGetClassName,
         defaultSelectorHandlers,
+        reset,
+        resetInjected,
     };
 }
--- a/src/inject.js
+++ b/src/inject.js
@@ -203,9 +203,13 @@ export const reset = () => {
     styleTag = null;
 };
 
+export const resetInjected = (key /* : string */) => {
+  delete alreadyInjected[key];
+};
+
 export const getBufferedStyles = () => {
     return injectionBuffer;
-}
+};
 
 export const startBuffering = () => {
     if (isBuffering) {
--- a/src/index.js
+++ b/src/index.js
@@ -13,6 +13,8 @@ const {
     flushToStyleTag,
     injectAndGetClassName,
     defaultSelectorHandlers,
+    reset,
+    resetInjected,
 } = Aphrodite;
 
 export {
@@ -24,4 +26,6 @@ export {
     flushToStyleTag,
     injectAndGetClassName,
     defaultSelectorHandlers,
+    reset,
+    resetInjected,
 };
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -38,7 +38,7 @@ export interface StyleSheetStatic {
      * Rehydrate class names from server renderer
      */
     rehydrate(renderedClassNames: string[]): void;
-    
+
     extend(extensions: Extension[]): Exports;
 }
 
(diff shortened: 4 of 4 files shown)


milesj:
@lencioni You have any feedback?

src/inject.js
```diff
@@ -203,9 +203,13 @@ export const reset = () => {
     styleTag = null;
 };
 
+export const resetInjected = (key /* : string */) => {
```
lencioni: I expected this function to reset all injected styles, not one specific style. Perhaps a better name would be `resetInjectedStyle` or something that tells consumers it only resets a single one?

dist/aphrodite.js
```diff
@@ -1400,7 +1400,7 @@ function hyphenateStyleName(name) {
 }
 
 var hyphenateStyleName$1 = /*#__PURE__*/Object.freeze({
-  default: hyphenateStyleName
+  'default': hyphenateStyleName
```
lencioni: Please revert all the changes in dist, that will be updated as part of the publishing step.

src/exports.js
```diff
@@ -188,5 +189,7 @@ export default function makeExports(
         flushToStyleTag,
         injectAndGetClassName,
         defaultSelectorHandlers,
+        reset,
+        resetInjected,
```
lencioni: Can you add documentation for these methods to the readme?

src/inject.js
milesj: Done!

dist/aphrodite.js
milesj: Done

src/exports.js
milesj: Added

README.md
```diff
@@ -130,6 +130,24 @@ const styles = StyleSheet.create({
 });
 ```
 
+## Resetting Style Cache
+
+The `reset` function can be used to reset the HTML style tag, injection buffer, and injected cache. Useful when Aphrodite needs to be teared down and set back up.
```
lencioni:
```suggestion
The `reset` function can be used to reset the HTML style tag, injection buffer, and injected cache. Useful when Aphrodite needs to be torn down and set back up.
```

src/inject.js
```diff
@@ -203,9 +203,13 @@ export const reset = () => {
     styleTag = null;
 };
 
+export const resetInjectedStyle = (key /* : string */) => {
+    delete alreadyInjected[key];
```
lencioni: Would it be sufficient to do this? 

```js
alreadyInjected[key] = undefined;
```

or do we have code that relies on it being fully deleted?

milesj:
@lencioni Not sure how to test this since `alreadyInjected` isn't exported, and exporting a `let` var doesn't seem like a good idea.

lencioni:
@milesj how about an integration test that uses aphrodite to inject styles into the DOM, and then removes the styles, and then calls your new function, and then uses aphrodite to inject that style into the DOM again and verify that all worked?

milesj:
@lencioni Added.

src/inject.js
milesj: We do an `Object.keys(alreadyInjected)`, so deleting seems like the best approach.

lencioni:
Thanks for the test! Do you know why CI is failing on Node 6?

milesj:
@lencioni I do not but will look into it.

I'd also suggest just dropping v6 since it's LTS is over.

milesj:
@lencioni @kevinbarabash Any update here?

lencioni:
Thanks for the ping! I'd really like for someone from Khan to review this. @kevinbarabash can you help with that?

milesj:
Bump? @kevinbarabash

PR #390, opened 2020-03-24
Async + parallel rendering

srolel:
This includes wrapping existing functionality in stateful classes, rather than having everything share the same global state. The API remains the same and all existing unit tests remain and pass.

In addition, `renderStaticAsync` is added which does what it sounds like, call a provided, async `renderFn` and return a promise with the rendered html and css.

Having stateful classes means we can instantiate them per request in the SSR use-case, this enabling parallel rendering.

note: PR into 2.4.0 since master is lagging behind it

### DIFF
--- a/src/exports.js
+++ b/src/exports.js
@@ -1,17 +1,7 @@
 /* @flow */
-import {hashString} from './util';
-import {
-    injectAndGetClassName,
-    reset,
-    resetInjectedStyle,
-    startBuffering,
-    flushToString,
-    flushToStyleTag,
-    addRenderedClassNames,
-    getRenderedClassNames,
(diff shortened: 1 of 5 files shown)


srolel:
[clabot:check]

ljharb:
Why was this closed and the branch deleted?

jlfwong:
Oh, huh -- I think it was closed automatically because I merged the PR that was the 2.4.0 branch.

The branch that was deleted isn't the branch for this PR -- it was the branch this PR was targeting, which was from #382.

I can't reopen it because the branch targeted by this PR no longer exists.

ljharb:
ah, yeah that would do it.

PR #405, opened 2025-07-24
Stop buffering style injection in `injectGeneratedCSSOnce`

timmcca-be:
## Summary:
This delay can lead to styles being used by consumers before they have been injected into the page. In partciular, it seems that changing focus interrupts the `asap` callback, and can delay it enough to cause visible issues. This is most apparent with a long `transition`, where you can see an element snap to its browser native styling before the correct styles are slowly applied.

I worry that this change, as implemented, will lead to functional and/or performance issues elsewhere. I'd appreciate some guidance around how best to test this, and how we may be able to resolve this more elegantly.

I'm also not sure why the build code changed so much. I had to run `npm i --legacy-peer-deps` to bypass some peer dependency conflicts, and the lockfile isn't committed in this repo ([this commit removed the lockfile](https://github.com/Khan/aphrodite/commit/ee1a58059cbca52f31527670e9ae79d655c16484), and its reasoning - "Since this repo is non-deployable, we don't want a lockfile." - does not make sense to me), so I suspect these changes are related to some inconsistent dependencies.

## Test plan:
Click the "Turn on" button on the examples page. Doing so without this fix will lead to the button briefly turning gray the first time. With this fix, the background will stay pink the whole time.


timmcca-be:
We have nobody who can review this or provide input on what downstream issues it may cause :(

xymostech:
Not sure who you were expecting to give review, but I saw this and thought I'd add my 2c!

I am quite confident that this will cause performance issues! Specifically, on a single render it will end up calling `flushToStyleTag()` many many times (one for each call to `css()`) instead of a single one at the end of the render. That was the point of the buffering in the first place. Each time you add new styles to the `<style>` tag, the browser has to do some thinking, and so doing one big modification is better than doing it 

If it's being slow because of `asap()`, then maybe switching to a different library than `asap` makes sense? That said, I think that the behavior that you are seeing in the examples is what I would expect, regardless of how slow `asap()` is being. You will have one paint with the original styles, the next paint will use the new classname but the styles likely won't have been injected yet, and a third paint when the new styles have been added. I would have to think about how to make that not happen. 🤔

If you were to get rid of the buffering (which I wouldn't recommend) then you could also get rid of the `flushToStyleTag()` export because nobody would need that any more. (although that would be a breaking change) And you can get rid of the bit in the readme.

